PR #50, opened 2023-07-03
Fix missing prefixes in example variant descriptions.

ifokkema:
Fixed missing prefixes in example variant descriptions.
Also, fixed three insertion descriptions that required the use of square brackets, but had none.


ifokkema:
@reece I took the liberty of adding a new label; "documentation fix". The other labels focus mostly on proposals, but I reckon we will have lots of PRs like this, simply correcting mistakes in the documentation.

docs/recommendations/general.md
```diff
@@ -56,7 +56,7 @@ In HGVS nomenclature some **characters** have a **specific meaning**
     * "**<font color="red">;</font>**" (semi colon) is used to separate variants and alleles; g.[123456A>G;345678G>C] or g.[123456A>G];[345678G>C]
     * "**<font color="red">,</font>**" (comma) is used to separate different transcripts/proteins derived from one allele; r.[123a>u, 122\_154del]
     * NC\_000002.11:g.48031621\_48031622ins[TAT;48026961\_48027223;GGC]
-    * NC\_000002.11:g.47643464\_47643465ins[NC\_000022.10:35788169\_35788352]
+    * NC\_000002.11:g.47643464\_47643465ins[NC\_000022.10:g.35788169\_35788352]
```
jfjlaros, line 59: Since this example is part of a definition of some sort, it would be good to check whether the omission of a numbering scheme is no longer allowed.

I seem to remember that not using a numbering scheme  was allowed in some cases and that counting started at the first nucleotide of the reference sequence in these cases. I can imagine that this example is a remnant of that practice. Perhaps @jtdendunnen can confirm that this practice is no longer recommended?

jfjlaros (commented):
All proposed changes look good to me, but there is one change that I would like to have checked.

docs/recommendations/general.md
ifokkema: > Since this example is part of a definition of some sort, it would be good to check whether the omission of a numbering scheme is no longer allowed.

The website is clear in my opinion; the prefix is mandatory:
> a letter prefix is mandatory to indicate the type of reference sequence used.

[source](https://hgvs-nomenclature.org/en/latest/recommendations/general/)

The number of examples missing a prefix is also almost negligible, common for typos in variant descriptions. This variant description is an example of the definition of the use of square brackets, so not a definition related to the prefix.


> I seem to remember that not using a numbering scheme was allowed in some cases and that counting started at the first nucleotide of the reference sequence in these cases.

I remember the `n.` prefix was used for that. I only recently removed code from LOVD that allowed the use of the `n.` prefix in `NM` reference sequences, indeed counting from the first base of the reference sequence. But Johan mentioned it wasn't allowed, and I couldn't find it anymore on the HGVS varnomen website, so I removed the code.

ifokkema: @jfjlaros I just had a meeting with @jtdendunnen today and asked him about this; he agrees that, indeed, in the past, the prefix was optional for an `ins[refseq:position]` description, then assuming the same prefix was used as the actual variant description. However, it was made mandatory a long time ago to match full variant descriptions, and all examples on the website should have been updated. He must have missed this one. Therefore, `NC_000002.11:g.47643464_47643465ins[NC_000022.10:35788169_35788352]` is incorrect and should be `NC_000002.11:g.47643464_47643465ins[NC_000022.10:g.35788169_35788352]`.

ifokkema:
@reece Do you have any comments on this PR, or can it be merged?

reece:
Yikes! Sorry about not seeing this @jfjlaros and @ifokkema -- it came in while I was on vacation in Europe. Looking now.

reece (approved):
Looks good to me. Thanks for finding and fixing these errors.

PR #57, opened 2023-08-16
Fix errors introduced in #54, and more.

ifokkema:
Pull request #54 introduced a few errors. I fixed those, and ran into more/similar errors with examples, and fixed those, too.
- Correct typos introduced in #54.
- Fixed several issues in an answer on the DNA repeat page.
  - Clarified "TG11" and "T7" to "11 TG copies" and "7 T copies", matching also the question asked.
  - Removed backslashes before underscores that are rendered in my Markdown preview, but otherwise, at least they are unnecessary.
  - Fixed the last variant description in the sentence, which was forgotten.
  - Ranges of repeats should be surrounded by parentheses to distinguish these from ranges of positions (based on the other examples on the website, and personal communication with Johan).
- Fixed several issues with an answer on the allele page.
  - Spaces aren't used in allele descriptions, so remove them in this example.
  - Repeat syntax requires square brackets, but they were forgotten in two places.
  - Repeat length ranges require parentheses.


reece (approved):
lgtm Thanks Ivo.

PR #70, opened 2023-10-11
Update the RNA syntax blocks

ifokkema:
Restructured all the syntax blocks of the RNA pages, according to the example from `main`.
- The syntax is laid out in tables with different rows in case of multiple syntax categories (e.g., on the alleles page).
- The syntax and the examples now include the sequence id / reference sequence. This clarifies that each variant description requires a reference sequence.
- Variables are defined as generic as possible, to allow later definitions to exist elsewhere.
- Fixed strings, like variant types ("del", "ins", etc) are surrounded by quotes.
  - Fixed strings that were previously not explained (like the "." between the coordinate type and the positions) are now also not explained.


reece (commented):
Overall, I think that this is looking very good.

Here are some discrepancies that we should discuss and resolve in our respective branches:

- I used `position_edit` as a general grammar elements; you used `first_change` and `second_change`, which are not grammar elements, but perhaps help comprehension
- I didn't describe literal text (like `"["` or `"del"`), but you did
 
Here's a representative example of the differences:

image

ifokkema:
Thanks for the review, @reece !

> * I used `position_edit` as a general grammar elements; you used `first_change` and `second_change`, which are not grammar elements, but perhaps help comprehension

I saw the `position_edit` definition, indeed, but I find this a confusing term. Or perhaps just "unknown", in the sense that, if you'd ask ten people what that field should contain, you'd likely get lots of different answers. Especially if people would have to click on a link to see the definition of the field, the definition would be a bit hidden, as well. Since variants are often referred to as sequence changes, I opted for "change" rather than "edit". It's not a full variant because the coordinate type is missing, so "variant" wouldn't be correct, either. Finally, it's not a position, so I decided to go for `first_change` and `second_change` even though this may suggest an order of things or a limitation of two changes only.


> * I didn't describe literal text (like `"["` or `"del"`), but you did

I decided to keep the changes limited to the change in markup. So I decided not to change the contents. Everything that was defined on the original pages, I defined in the updated syntax blocks as well. What wasn't defined, I didn't add.


> In this case, I'd vote to eliminate all syntax descriptions except for the link to the grammar explanation (which is currently incomplete)

I think we should run this by the group. I noticed that in the IDE it takes a few seconds only to check if all definitions are the same on all pages, so it's not a lot of overhead to redefine them every time. Another possibility would be to fetch the definitions from elsewhere, so they're shown on each page, but they don't need to be redefined in all files.

reece:
Thanks @ifokkema . I'll run options by the rest of the committee using your examples.

reece:
Merged branch into 77-make-syntax-and-explanatory-text-consistent-across-dna-rna-aa to create consistency across efforts.
Closing issue and branch.

PR #71, opened 2023-10-15
61 update the syntax blocks dna pages

reece:


docs/recommendations/DNA/alleles.md
```diff
@@ -76,32 +89,32 @@ Allele: a series of variants on one chromosome.
...
+!!! note "I have a patient with hearing loss and variants in the GJB2 (`c.35delG`) and GJB6 (`c.689*690insT`) genes, how should I describe this? \*(Nancy Carson, Ottawa, Canada)\_"
 
-    The recommendation is to use the format GJB2:c.[35delG] GJB6:c.[689_690insT]. This uses standard HGVS descriptions and prevents confusion regarding which variant was found in which gene. Note it is essential that you also define the coding DNA reference sequence used. Another format, coping with this directly, is to describe the variants as NM_004004.2:c.[35delG] NM_006783.1:c.[689_690insT], i.e. using the Genbank reference sequences in stead of the HGNC approved Gene Symbol.
+    The recommendation is to use the format `GJB2:c.[35delG] GJB6:c.[689_690insT]``. This uses standard HGVS descriptions and prevents confusion regarding which variant was found in which gene. Note it is essential that you also define the coding DNA reference sequence used. Another format, coping with this directly, is to describe the variants as `NM_004004.2:c.[35delG] NM_006783.1:c.[689_690insT]`, i.e. using the Genbank reference sequences in stead of the HGNC approved Gene Symbol.
```
ifokkema, line 102: I haven't looked in detail yet at the changes in the discussion section, but I noticed there's a backtick too many after the first variant description.

docs/recommendations/DNA/inversion.md
```diff
@@ -4,11 +4,23 @@
...
+<td><code>sequence_identifier ":" coordinate_type "." position_range "inv"</code></td>
+</tr>
+<tr>
+<th>Examples</th>
+<td><code>NC_000001.11:g.123_129inv</code></td>
+</tr>
+</table>
+
+- `sequence_identifier` = an identifier for DNA sequence
+- `position_range` = a single nucleotide position or inclusive range of deleted sequence
```
ifokkema, line 21: This definition of the `position_range` isn't correct; a single position is excluded.

docs/recommendations/DNA/repeated.md
```diff
@@ -4,32 +4,34 @@
...
+</tr>
+<tr>
+<th>Examples</th>
+<td><code>NC_000001.11:g.[123G>A];[345del]</code></td>
+</tr>
+</table>
+
+- `position` = the first (5'-most) nucleotide of the repeat
+- `sequence` = the sequence of the repeat
+- `number` = the number of occurances of the repeat
```
ifokkema, line 33: I used `copy_number` here, to specify it a bit. It's also not always just a number; it can also be a range: `[(10_20)]` is valid syntax for an uncertain length.

docs/recommendations/DNA/substitution.md
```diff
@@ -4,17 +4,24 @@
...
+<td><code>sequence_identifier ":" coordinate_type "." position reference_nucleotide ">" alternate_nucleotide</code></td>
+</tr>
+<tr>
+<th>Examples</th>
+<td><code>NC_000001.11:g.123A>C</code></td>
+</tr>
+</table>
+
+- `sequence_identifier` = an identifier for DNA sequence
+- `position` = a single nucleotide position or inclusive range of deleted sequence
```
ifokkema, line 21: This can not contain a range, but it has to be a single position. I struggled with the definitions and decided to go for `position` here while I used `positions` where variants can have a single position as well as a range. However, I used `positions` as well for insertions and inversions, that _must_ have a range. I like your separate definition for that, although `position_point` may suggest it's one position only. How about `interbase_position` for insertions and `position_range` for inversions? And, possibly, `single_position` for substitutions?

docs/recommendations/DNA/alleles.md
```diff
@@ -30,14 +30,17 @@ Allele: a series of variants on one chromosome.
...
-<td><code>accession ":" type "." "[" posedit "(;)" posedit"]"</code></td>
+<td><code>sequence_identifier ":" coordinate_type "." "[" position_edit "(;)" position_edit"]"</code></td>
 </tr>
 <tr>
 <th>Examples</th>
 <td><code>NC_000001.11:g.[123G>A(;)345del]</code></td>
 </tr>
 </table>
 
+- `position_edit` = any g. position-edit, such as [deletion](./deletion.md), [insertion](./insertion.md), or [deletion-insertion](./delins.md)
```
ifokkema, line 41: It's not necessarily `g.` based; this page is also used for the `c.` coordinate type.

ifokkema (commented):
It's great that we can compare this way; it clearly highlights what still needs to be defined. Besides the points already brought up by you, I noticed:
- I used the parts from the examples in the definitions, like the original pages do, too. E.g. `<code>sequence_identifier</code> = the sequence identifier used; NM_004006.3`. I find this a bit easier to understand, but it requires the (re)definition of each term on each page. So it probably needs discussion within the group.
- The terms related to positions. `position`, `positions`, `position_range`, `position_point`... we need to synchronize these definitions.
- Some small typos or issues that can easily be fixed; I highlighted them as separate comments.

docs/recommendations/DNA/repeated.md
reece: Okay. I'm finding it hard to know how to take comments about content when I'm working on process.  Let's please defer proofreading a little bit. I appreciate your careful eye!

ifokkema: Alright!

reece:
Merged into #77.

Issue #63, opened 2023-10-02
Update the syntax blocks; protein pages


ifokkema:
## Update the syntax blocks; protein pages
- Update all syntax blocks based on the new format as agreed upon on Oct 2nd, to be committed in branch `23-create-and-use-style-guide`.
- Separate issues are created for each category of nomenclature pages. This issue is for the protein pages.

### Details
- Related to #23.
- Create a new branch for this issue, and a separate PR, to be merged into `main`.
- Branch of off [23-create-and-use-style-guide](https://github.com/HGVSnomenclature/hgvs-nomenclature/tree/23-create-and-use-style-guide), or, if that has been merged into `main` already, branch of off [main](https://github.com/HGVSnomenclature/hgvs-nomenclature/tree/main).
- To keep handling the PRs as simple as possible, stick to only redesigning the syntax blocks. Other changes can be committed in other branches.

reece:
Merged branch into 77-make-syntax-and-explanatory-text-consistent-across-dna-rna-aa to create consistency across efforts.
Closing issue and branch.

PR #73, opened 2023-10-15
72 move syntax specifications to yaml

reece:
This PR is a proposal to move the syntax blocks from specific recommendation files to a consolidated file with definitions. The advantages are:

- moves toward treating the syntax as source (rather than just text in markdown)
- helps to further standardize the structure of presentation
- enables new presentations, such as a matrix of {dna,rna,aa} ⊗ variant types. 

There is no change to the docs build process.

This screenshot shows the rendering:
image

This PR does *not* update recommendations. Instead, if we choose to merge this PR, we'll merge main into existing change branches (#70 , #71) and adapt there.

P.S. Please ignore the posedit and other syntax details; they are irrelevant for this PR.

### DIFF
--- a/docs/recommendations/DNA/alleles.md
+++ b/docs/recommendations/DNA/alleles.md
@@ -6,9 +6,48 @@ Allele: a series of variants on one chromosome.
 
 ## Syntax
 
+<table class="syntax">
+<tr><th class="section" colspan=2>Variants in <i>cis</i></th></tr>
+<tr>
+<th>Syntax</th>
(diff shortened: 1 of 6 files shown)


reece:
Here's another view with proteins from #63 : 
image

docs/recommendations/DNA/alleles.md
```diff
@@ -6,37 +6,9 @@ Allele: a series of variants on one chromosome.
 
 ## Syntax
 
-<table class="syntax">
```
ifokkema, line 9: The loss of these examples is still unintended, right?

reece: Yes, for this round. Fixed.

PR #74, opened 2023-10-15
63 update the syntax blocks protein pages

reece:
This PR converts all of the old Description blocks to Syntax blocks using the syntax.yaml file approach (#73). (If we decide that to not use the syntax.yaml file approach, the HTML tables are easily copied.)

This PR also includes some updates to the `pull-syntax` script to also permit markdown-formatted notes.


reece:
Merged into #77 .

PR #75, opened 2023-10-20
Update the formatting of the RNA pages.

ifokkema:
## Update the formatting of the RNA pages.
I found several issues during this reformatting.
To be decided:
- Should parts of variant descriptions, such as coordinate types, positions, bases or short sequences, and variant types ("del", "delins", etc) also be quoted?
- The original pages have a better markup; example, newline, more info, newline, possible NOTE. In our current rendering, this is all one long line. I find the original markup more readable, but I believe it requires `<br>` tags to achieve this in Markdown.

Still to do:
- The notes sometimes refer to the variables used in the syntax description. Because these may still change, I have not updated these. This still needs to be done.
- Given sequences that describe the context of a variant may sometimes need a legend. When I thought it would improve readability, I kept both sequences given ("before" and "after") and highlighted the sequence that would be deleted in the "before" sequence and the replaced sequence in the "after" sequence. For instance, on the RNA dup page, the last comment.

Nomenclature issues:
- The RNA insertion page contains faulty suggestions. Same for the RNA inversion page. I have not removed them, as we're currently unsure how to describe this. I did mark them as invalid.
- I find the last suggestion on the RNA delins page against the standards (splitting a variant up into two descriptions because one part is a common variant). But it's similar to the exception given on the delins page that is an exception to the exception as described on the RNA substitution page (two substitutions become one delins if they affect the same codon). This is confusing, and it would be good to straighten this out.

Technical issues:
- The indentation present in the original pages is often lost. Lists in lists are rendered as a single list. This may be an issue with our rendering because my Markdown preview works, but `mkdocs` doesn't render it well unless I increase the indentation of the sublist items in the source code. But I can't commit that, because the pre-commit hook keeps undoing my changes. See, for instance, the RNA delins examples. There seems to be one sublist, but there should be more.

### DIFF
--- a/docs/recommendations/RNA/deletion.md
+++ b/docs/recommendations/RNA/deletion.md
@@ -22,25 +22,25 @@ bin/pull-syntax -f docs/syntax.yaml rna.deletion
 - the "position(s)\_deleted" should contain **two different positions**, e.g. 123_126 but not 123_123.
 - the "position(s)\_deleted" should be listed from **5' to 3'**, e.g. 123_126 but not 126_123.
 - for all descriptions the **most 3' position** possible of the reference sequence is arbitrarily assigned to have been changed (**3'rule**)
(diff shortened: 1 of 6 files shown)


reece:
Merged branch into 77-make-syntax-and-explanatory-text-consistent-across-dna-rna-aa to create consistency across efforts.
Closing issue and branch.

PR #92, opened 2023-11-20
48 create page for known tools

reece:
This PR implements a list of software for the HGVS Nomenclature pages. The preview is available at 
https://hgvs-nomenclature.org/en/48-create-page-for-known-tools/software/ .

@Peter-J-Freeman and @jfjlaros : I took the liberty of making entries for VV and Mutalyzer using descriptions from github. Feel free to edit the source files (in docs/software/).

I would also appreciate your comments on the scope of the software pages described at the top of the preview page. Specifically:
- open source only
- general purpose tools that manipulate user-specified variants rather than annotation tools that happen to use HGVS
- type and functionality categories


jtdendunnen:
The page looks like a good start. VV versus Mutalyzer seems a bit out of balance, the VV text reads a bit like a commercial. Did Jeroen alreay check the Mutlayzer description?

I think NCBI has build some kind of package as well, correct?

The UCSC browser is able to use HGVS variant descriptions and jump to the variant in the genome; should this be on the list as well?

reece:
@jtdendunnen Yes, I think they're out of balance. I pulled text nearly verbatim from the README files in each repo asked Peter and Ivo to update with text that they're comfortable with before this goes live.

NCBI had a tool called [Remap](https://www.ncbi.nlm.nih.gov/genome/tools/remap), but it's been retired.

Do you agree with the principles described in https://hgvs-nomenclature.org/en/48-create-page-for-known-tools/software/#submissions, which says:

> The software must:
> 
> - Be *primarily* concerned with provided the functionality listed above.
>   Software that merely accepts or emits HGVS expressions as part of their
>   operation (e.g., variant annotation) do not meet this criterion.
> - Meet the license requirements above. Commercial software, including hybrid
>   open-source/commercial software, will not be included.
> - Have a publication in PubMed. This requirement is intended to create a bar
>   that implies review by peers.

murphyte:
NCBI has a suite of APIs described at: https://api.ncbi.nlm.nih.gov/variation/v0/
I'll see if I can get someone to provide some suitable text.

docs/software.md
```diff
@@ -0,0 +1,61 @@
...
+
+The software must:
+
+- Be *primarily* concerned with provided the functionality listed above.
+  Software that merely accepts or emits HGVS expressions as part of their
+  operation (e.g., variant annotation) do not meet this criterion.
+- Meet the license requirements above. Commercial software, including hybrid
+  open-source/commercial software, will not be included.
+- Have a publication in PubMed. This requirement is intended to create a bar
+  that implies review by peers.
```
ahwagner, line 50: This feels a little artificial. What does it mean to have a PubMed ID? How central a piece of software is to an article can be highly variable. Will we make that distinction? If so, are we effectively becoming peer review? And if that, why require an article at all?

I think it would be better for us to define specific criteria relating to claimed functionality. Perhaps require that any implementations supporting one or more of the above functions pass some standard functionality test for each such function?

docs/software.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **Functionality**: Software is annotated as performing one or more of the following functions:
+    - conversion: Does the software convert to/from HGVS and other formats (e.g., VCF, SPDI, VRS)
+    - liftover: Does the software provide *direct* liftover between aligned genomic sequences?
+    - normalization: Does the software normalize variants (e.g., 3' shift) and rewrite (e.g., `ins` → `dup`)
+    - parsing: Does the software parse HGVS expressions into a usable structure for analysis?
+    - projection: Does the software project variants from one sequence to another?
+    - translation: Does the software infer protein consequences from transcript variants?
+    - validation: Does the software provide syntactic validation (i.e., the correct form) and semantic validation (i.e., the parsed structure makes sense)?
+
```
ahwagner, line 37: We should provide criteria around what belongs in the `description` section, if we are to have one. This will help keep software descriptions balanced.

lib/osi-licenses.yml
```diff
@@ -0,0 +1,54 @@
+# Generated by ChatGPT on 2023-11-18
+# https://chat.openai.com/c/9785e91f-37ab-4d13-8221-8e1bb10e8bc1
```
ahwagner, line 2: This URL is not accessible without a login. I did not test if it is accessible with a login.

lib/osi-licenses.yml
```diff
@@ -0,0 +1,54 @@
+# Generated by ChatGPT on 2023-11-18
```
ahwagner, line 1: This provenance is helpful!

lib/osi-licenses.yml
```diff
@@ -0,0 +1,54 @@
...
+  name: MIT License
+  url: https://opensource.org/licenses/MIT
+
+MPL-2.0:
+  name: Mozilla Public License 2.0
+  url: https://opensource.org/licenses/MPL-2.0
+
+OTHER:
+  name: Other/uncommon license
+  url: https://opensource.org/licenses/?categories=popular-strong-community
```
ahwagner, line 54:
```suggestion
```

I know ChatGPT came up with this list, but `OTHER` is out of place here. URL is definitely not to some sort of definition about what constitutes an `OTHER` license. Recommend removing.

ahwagner (changes requested):
Overall, really like this addition. A few nitpicky change requests follow in this review.

docs/software.md
reece: Okay, maybe that's the easiest thing to do right now. I'm looking for easy wins that don't create lots of debate and deliberation. So, I'll remove the description column.

docs/software.md
reece: Well, kinda like it say... it's a bar. I'd like to stay away from debates about functionality until we have a compliance suite. In the absence of that, I think it needs to be self-described.

lib/osi-licenses.yml
reece: I think it's critical for this list to have a set of principles, one of which is to use OSI's definition of common licenses.  I didn't want to add every niche license.

Unfortunately, when I added VV I learned that it uses the Affero license. That necessited an OTHER category, which I added manually.

lib/osi-licenses.yml
reece: Well, it really came from https://opensource.org/licenses/?categories=popular-strong-community, which I had ChatGPT summarize for me as yaml. So that OSI link is a better reference. This was really a note to myself.

reece:
> NCBI has a suite of APIs described at: https://api.ncbi.nlm.nih.gov/variation/v0/ I'll see if I can get someone to provide some suitable text.

@murphyte: Ooh, I forgot about that. I added a preliminary entry for it. And because @jtdendunnen and @ahwagner expressed concern about bias in the descriptions, I dropped the column, so that bit isn't needed.

The result is at https://hgvs-nomenclature.org/en/48-create-page-for-known-tools/software/.

jfjlaros:
Since the list contains a mix of programming libraries and tools, I presume the Mutalyzer libraries (see [previous discussion](https://github.com/HGVSnomenclature/hgvs-nomenclature/discussions/56#discussioncomment-6662191)) should be added as well?

The metrics (number of stars, forks, etc.) are misleading at best, e.g., for Mutalyzer we only report the number of stars on one (relatively new) repository, while the project as a whole counts 161 stars in total. My suggestion would be to remove these metrics completely.

The definition of "liftover" seems to be recursive, I am not sure what it means.

Some suggestions for functionality tags:
- transcription: Does the software infer transcript variants from genomic variant descriptions?
- backtranslation: Does the software infer possible transcript variants from observed protein consequences?
- extraction: Is the software able to generate a description from a reference and observed sequence?
- allele-aware: Does the software handle allele descriptions?
- ensembl: Does the software handle Ensembl reference identifiers (e.g., ENSG, ENST)?
- rna: Does the software handle RNA variant descriptions?
- protein: Does the software handle protein descriptions?

reece:
@ifokkema : The current Mutalyzer entry was a sketch by me. As I suggested at the top of the PR, please edit when you're able (I think that you're still on your honeymoon). I chose the repo I did because the relationship of tools was not obvious to me. Please choose whatever you like, including deciding whether you want to split or merge the mutalyzer entries. BTW, the new (to me) Mutalyzer 3 site is very nice looking.

Peter-J-Freeman:
@reece @ifokkema. Will update the VV entry. Do we have a final schema, or are we playing a little liberally? Just want to check before I update

reece:
> @reece @ifokkema. Will update the VV entry. Do we have a final schema, or are we playing a little liberally? Just want to check before I update

The schema is pretty stable. My preference is to merge this more or less as-is, then come back for a second pass to consider new functionality categories, whether to bring descriptions back and whether to have criteria for them, and address other comments here.

reece:
@Peter-J-Freeman : I'd love to know your thoughts about the license options.

Assembling a list like this tends to create tensions about what software to include and how to describe it and who gets to choose. So, I started out by articulating principles rather than making ad hoc decisions. 

One of the principles was to limit the list of licenses to those with "strong community support" by OSI. IMO, preventing license sprawl is a worthwhile mission, like FAIR principles, code coverage, or anything that makes for good practice.

However, Affero isn't a popular license so the VV entry shows OTHER.

If you're okay with that, great. If not, I'd appreciate suggestions for how to modify the principles while still keeping the objective to limit license sprawl.

Peter-J-Freeman:
@reece. Happy to go with other. The licence choise was the university's decision. Mutalyzer used GPL 3 too. I'm not too worried about putthig other since the License is stated in our repos

My main question is around the format. I think it might be worth splitting VariantValidator from the VariantValidator API otherwise we will need too much granularity. This is a mock example which would be ideal and really useful for users, especially until I get the time to write an update paper. 

```yaml
name: VariantValidator
primary_url: https://variantvalidator.org/
source:
  type: github
  repo_slug: openvar/variantvalidator
types:
- web api
- web user interface
- python libraries
- docker builds
functionality:
- conversion
- normalization
- parsing
- projection
- translation
- validation
- liftover
license: AGPL-3.0 license
website: 
  url: https://variantvalidator.org/
  source:
    type: github
    repo_slug: openvar/variantvalidator
  tools:
    Validator:
      description:
      - Validation of multiple sequence variant description formats
      inputs:
      - vcf
      - hgvs g. c. r. p.
      - other
      - RefSeq Data
      outputs:
      - vcf
      - hgvs g. c. r. g.
      - Gene information
      - links to UCSC genome browser, gnomAD, Varsome, RefSeq
      - liftover
      - RefSeq data
    GenesToTranscripts:
      description:
      - Shows available transcripts for a gene
      inputs:
      - HGNC:ID
      - gene symbol
      outputs:
      - gene symbol
      - gene id
      - available transcripts
      - annotation e.g. Mane Select
      - RefSeq data
    BatchValidator:
      description: 
      - VariantValidator functionality for multiple descriptions
      inputs:
      - see VariantValidator
      outputs:
      - see VariantValidator
api: 
  url: https://rest.variantvalidator.org/
  source:
    type: github
    repo_slug: openvar/rest_variantValidator
  tools:
    VariantValidator:
      description:
      - Validation of multiple sequence variant description formats
      inputs:
      - vcf
      - hgvs g. c. r. p.
      - other
      - Refseq Data
      outputs:
      - vcf
      - hgvs g. c. r. p.
      - gene information
      - lifover
      - Refseq data
    LOVD:
      description:
      - Genomic description to transcript and protein descriptions for use in pipelines.
      inputs:
      - vcf
      - hgvs g.
      outputs:
      - vcf
      - hgvs g. c. p.
      - gene information
      - liftover
      - RefSeq desctiptions
      - Ensembl descriptions
    gene2transcripts_v2:
      description:
      - Shows available transcripts for a gene, alignment data and information about each transcript
      inputs:
      - HGNC:ID
      - Gene symbol
      outputs:
      - gene id
      - gene symbol
      - transcript information
      - alignment data
      - Ensembl data
      - RefSeq data
    hgvs2reference:
    - Accepts a HGVS formatted description and returns the reference bases and alt bases
    inputs:
    - hgvs g. c.
    outputs:
    - reference bases
    - alt bases
    - normalized descriptions
    
description: |
  VariantValidator is a user-friendly software tool designed to validate the
  syntax and parameters of DNA variant descriptions according to the HGVS
  Sequence Variant Nomenclature.

  VariantValidator ensures that users are guided through the intricacies of the
  HGVS nomenclature, e.g. if the user makes a mistake, VariantValidator
  automatically corrects the mistake if it can, or provides helpful guidance if
  it cannot. In addition, VariantValidator accurately inter-converts between
  transcript variant descriptions and genomic variant descriptions in HGVS and
  Variant Call Format (VCF)

  VariantValidator interfaces with the hgvs package to parse, format, and
  manipulate biological sequence variants. See
  https://github.com/biocommons/hgvs/ for details of the hgvs package

  rest_VariantValidator and the Python libraries are highly functional platforms enabling high-throughput and
  embeddable utilisation of functionality of https://variantvalidator.org/

pubmed_id: "28967166"
```

What do you think is best??? Splitting or somethig like the mock above?

reece:
@Peter-J-Freeman Thanks for the quick turnaround.

For this PR, can we focus on changes within the current framework? In other words, I think you should limit your edits in this PR to the current structure.

You raise some good ideas for a subsequent version, but they're a big lift. And once the info becomes as dense as you propose, I think it's much better to just link to the project's home page. IMO, the success of this page depends on it being a useful pointer to information rather than a comprehensive review of packages. But, in any case, the best way IMO to test this is to merge the minimal version, then update with v2 and get opinions from the group.

Peter-J-Freeman:
@reece. Happy to keep Other for the License. The GPL3 wasn't my decision. I think Mutalyzer used to have it too @jfjlaros?

I have made edits and submitted a PR.

jfjlaros:
> I think Mutalyzer used to have it too

Mutalyzer 2 was released under the AGPL license indeed.

Peter-J-Freeman:
@jfjlaros I thought so. What are you on now? Funnily enough, I'm in the process of asking our translation team to review the license. @reece , thanks for the pointers too

jfjlaros:
> What are you on now?

I think we use the MIT license for Mutalyzer 3 and all the libraries it depends on. Some of the older libraries may still use other (or no) licenses.

Peter-J-Freeman:
@reece hope the PR is OK. Let me know if you want any additional changes before merge.

reece:
I'm fine with both of you creating separate entries for the API and web interface components if that's what you think is most useful for readers.

The only thing I ask is that we not fragment entries by specific functionality, which I think we should capture explicitly (and perhaps expand in a future version).

Peter-J-Freeman:
I'm happy with the single entry at the moment. Think it is only worth extracting if it will break something. Looking again at the structure, we can quite easily add links to docs pages for the API as well as interactive site and easily describe what each is in the current structure. Another job for the to-do list. Get round to making the docs!!! Really need it for the API above all else to be honest. I'm happy with my [PR](https://github.com/HGVSnomenclature/hgvs-nomenclature/pull/94) now. Don't think I need to add anything else for v1.

lib/osi-licenses.yml
reece: @ahwagner : My preference is what's listed here. However, here are two other options:

1) Keep the OTHER label, but make the link point to the primary_url (i.e., the user needs to go figure it out)
2) Keep the OTHER label, but don't link at all

If you like one of those more than what's there, I'm happy to make the change.

ahwagner: Okay, I think that's fine. I took another look, and this can be easily resolved by updating to the correct URL (which does exist, but the original URL does not point there).

```suggestion
OTHER:
  name: Other/uncommon license
  url: https://opensource.org/licenses/?categories=other-miscellaneous
```

ahwagner (approved):
[Made one change](https://github.com/HGVSnomenclature/hgvs-nomenclature/pull/92#discussion_r1405294391) to address [my prior comment](https://github.com/HGVSnomenclature/hgvs-nomenclature/pull/92#discussion_r1399991605) about OTHER URL without removing the URL entirely.

With that change this all looks good to me!

lib/osi-licenses.yml
reece: That's not actually the correct URL. OTHER in license.yml means that it's not one of the licenses that we choose to recognize. other-miscellaneous refers to licenses that don't fit the categories at OSI.

Example: Affero is used by VV, uncategorized at OSI, and listed as OTHER for us. Linking to other-miscellaneous is not correct.

PR #93, opened 2023-11-20
Update glossary.md

jtdendunnen:
please check my suggested update, seems clearer to me

### DIFF
--- a/docs/background/glossary.md
+++ b/docs/background/glossary.md
@@ -72,9 +72,9 @@ Please note that this Glossary is **work in progress**. Do you encounter missing
 
 - homozygous: an individual in which both alleles at a given locus are **identical** ([MESH](http://www.ncbi.nlm.nih.gov/mesh?term=homozygous)).
 
-- hypermorphic variant: a variant causing the gene product to possess an increased level of activity, or in which the normal (wild-type) gene product is expressed at an increased level
+- hypermorphic variant: a variant characterized by a partial gain of gene activity (including an increase in protein production or function)
 
-- hypomorphic variant: a variant causing the gene product to possess a reduced level of activity, or in which the normal (wild-type) gene product is expressed at a reduced level
+- hypomorphic variant: a variant characterized by a partial loss of gene activity (including a reduction in protein production or function)
 
 - indel: **HGVS**: confusing term, not used: **sometimes**: a sequence change where, compared to a reference sequence, one or more nucleotides are replaced by one or more other nucleotides: **sometimes**: a variant which is a deletion or an insertion.: **sometimes**: (evolutionary biology) a type of variant in which a specific nucleotide sequence is present (insertion) or absent (deletion). : [MESH](http://www.ncbi.nlm.nih.gov/mesh?term=indel): a length difference between two alleles where it is unknowable if the difference was originally caused by a sequence insertion or a sequence deletion
 


docs/background/glossary.md
```diff
@@ -72,9 +72,9 @@ Please note that this Glossary is **work in progress**. Do you encounter missing
 
 - homozygous: an individual in which both alleles at a given locus are **identical** ([MESH](http://www.ncbi.nlm.nih.gov/mesh?term=homozygous)).
 
-- hypermorphic variant: a variant causing the gene product to possess an increased level of activity, or in which the normal (wild-type) gene product is expressed at an increased level
+- hypermorphic variant: a variant characterized by a partial gain of gene activity (including an increase in protein production or function)
```
reece, line 75: How about:

- hypermorphic variant: a variant characterized by an increase of activity of the gene product, perhaps through increased expression, increased enzymatic activity, or increased stability.

And similarly for hypomorphic. 

I specifically avoided "protein" here because an increase in enzymatic of ribozymes, for example, should still qualify.

reece (commented):
@jtdendunnen See specific changes to definitions below.

But, a broader comment: In my opinion, HGVS Nomenclature should define only those terms that are required for the specification. We don't use "hypermorphic" or "hypomorphic" anywhere in the text, so my preference is to drop them.

docs/background/glossary.md
ahwagner: I agree with @reece about the broader notion of preserving only vocabulary that is used within the specification.

As for the specific definitions, I agree with the changes that @jtdendunnen have made as well as @reece's comment that we should avoid use of the word "protein". I would prefer the use of "gene product" instead. Reece also dropped the use of "partial gain" in favor of "increase" for _hypermorphic_; another change I agree with. There's no notion of a "complete gain" so the use of "partial" here is awkward. "partial loss" for _hypomorphic_ makes sense, since a "complete loss" would be _amorphic_.

Structurally, I prefer @jtdendunnen's use of parentheses in lieu of a clause beginning with "perhaps".

So, aligning these preferences with what is already written, I think I would define these as:

> hypermorphic variant: a variant characterized by a gain of gene activity (e.g., by increased gene product expression or function).

> hypomorphic variant: a variant characterized by a partial loss of gene activity (e.g., by decreased gene product expression or function).

ahwagner (changes requested):
I agree with @reece about the broader notion of preserving only vocabulary that is used within the specification.

I have also added some specific recommendations to the definitions if we intend to keep these.

reece:
I like Alex's change, but I still prefer dropping the definitions entirely.

docs/background/glossary.md
ahwagner: I'm not sure why, but my changes didn't seem to stick? I may have messed this up. In summary, like what @jtdendunnen and @reece did here, and would combine them to create the following:

_hypermorphic variant_: a variant characterized by a gain of gene activity (e.g., an increase in gene product expression or function).

_hypormorphic variant_: a variant characterized by a partial loss of gene activity (e.g., a decrease in gene product expression or function).

jtdendunnen:
Reece stated: We don't use "hypermorphic" or "hypomorphic" anywhere in the text, so my preference is to drop them.

I am mixed here. Specific HGVS definitions are provided at the specific pages, like deletion, duplication, etc. The Glossary is not a "HGVS definition's page", it is a page explaining what is meant with terms around the topic of "variants". Since I received questions about "variants", like "what is a hypomorphic variant" I added them to the Gloossary. Maybe a good we mention this specifically on the Glossary page and in addition clearly mark/highlight HGVS definitions.

ifokkema (approved):

reece (approved):
lgtm

PR #95, opened 2023-11-22
add (some) new functionality items recommended by @jfjlaros

reece:
@Peter-J-Freeman and @jfjlaros : I added some of the categories recommended by @jfjlaros and updated entries accordingly. 

@jfjlaros : I chose the categories that (to me) seemed like new functionality that was not represented, as opposed to constraints or commentary that was already covered.


reece:
Merging into origin/48-create-page-for-known-tools. We can pick up more conversation there if needed.

Issue #98, opened 2023-11-26
Check style for nested lists


ahwagner:
Differences between items at different levels of nested lists are too subtle to discern. It makes entries like [glossary:mutation](https://hgvs-nomenclature.org/en/latest/background/glossary/#:~:text=mutation%3A%20NOTE,a%20disease%20phenotype.) difficult to read.

ahwagner:
I looked into this; relevant background in [this](https://github.com/mkdocs/mkdocs/issues/545) issue.

I propose that–for the short term–we address this with the `mdx_truly_sane_lists` extension as discussed in that thread, which will help us render historical content correctly without much additional work.

Long term, it might make sense to update our linter to capture markdown intents that are not multiples of 4, and address older lists.

reece:
Merged in

PR #99, opened 2023-11-26
fix md lists without indent multiples of 4

ahwagner:
addresses #98

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -20,6 +20,7 @@ markdown_extensions:
 - markdown_captions
 - pymdownx.details
 - pymdownx.superfences
+- mdx_truly_sane_lists
 - toc:
     permalink: "#"
     toc_depth: 1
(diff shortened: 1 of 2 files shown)


reece:
Would you be open to just indenting 4 spaces everywhere? 4 spaces is the intended spec, even if followed poorly. And, it's also easier to read and consistent with most of the HGVS Nomenclature markdown.

Sigh... we ended up with the mix of indents because I was using prettier as a pre-commit hook. It was rewriting markdown lists to make them *two* spaces because that what it wants. In fact, the author is so insistent that prettier should NOT be configurable. So, that makes prettier incompatible with intended markdown AFAICT. My intention was to make it 4 spaces everywhere.

ahwagner:
@reece I agree. I suggested in https://github.com/HGVSnomenclature/hgvs-nomenclature/issues/98#issuecomment-1826463614 that we use this as a short-term work-around while we fix the indent issue; I also am leaning towards compliance with markdown convention.

This is a minimal effort solution and I won't lose any sleep over it if you want to close the PR and just go straight to adjusting the linter. Or we can merge this and push that effort out for another day if we have other more pressing issues.

reece:
Check out this alternative: https://github.com/HGVSnomenclature/hgvs-nomenclature/pull/101

reece:
I merged the indentation fix since that seemed to be the direction @ahwagner wanted to go also

PR #102, opened 2023-11-27
Update insertion.md

jtdendunnen:
corrected N[25], removed duplicated text


ifokkema (approved):
Looks good to me!

reece (approved):
Good catch! Looks good to me too.

PR #109, opened 2023-11-30
restore text that was indvertently deleted from DNA/alleles

reece:
Restore text that was inadvertently deleted from DNA/alleles. Also, fixes the syntax table and examples. 

@ifokkema @rileygr: This PR should address the text that was inadvertently removed from DNA/alleles. Does this solve the problem?


ifokkema (approved):
Looks good to me! I made one small change to the styling of a variant description in the notes.

reece:
Good catch, @ifokkema . Thanks.
(And we may have a lot more of the "don't write it like this" variants to style... might be worth an explicit pass)

Issue #114, opened 2023-12-01
Unclear definition of new repeat syntax


ifokkema:
The exact definition of the "new" repeat syntax (with multiple positions given) is used in a multitude of different ways in the current docs.

- **Entire range, sequence given.** E.g., `LRG_763t1:c.54_110GCA[23]` on the DNA repeat page. The range and sequence together indicate the reference sequence contains 19 repeats, so this is an extension.
- **Entire range, no sequence given.** E.g, `r.456_465[4]466_489[9]490_499[3]` on the RNA repeat page. The range is given, but since the sequence isn't given, I can only assume what the repeat unit length was. Anyway, the first and last given position lengths are 10 bases, which can't be a repeat, so this example is wrong either way.
- **First repeat only, no sequence given.** E.g., `r.123_125[23]` on the RNA repeat page. The positions only indicate the first repeat, so I know the repeat unit size, but can't tell whether this is an extension or a contraction unless I look up the reference sequence.

For sure, the third description is incompatible with the other two descriptions. But let's stick to only one of the above three options. Which one is correct?

jtdendunnen:
- "r.456_465[4]466_489[9]490_499[3]"
This format is not allowed, i.e. we go towards the format where region and sequence of the repat must be indicated. So far I did this for the DNA variant pages, not yet for the RNA pages. The RNA pages therefore contain incorrect examples. I suggest we chahnge these all. So correct would be something like r.456_465TA[4]466_489CAG[9]490_499GCCAG[3].

- "r.123_125[23]"
Similar, this is not correct should be like r.123_125CAG[23]

ifokkema:
@jtdendunnen 
I started fixing this in 1ff5fda836cdc65ec0000ef5c2b75b36766513df when I realized we do not have clarity on this.

> * "r.456_465[4]466_489[9]490_499[3]"
>   This format is not allowed, i.e. we go towards the format where region and sequence of the repat must be indicated. So far I did this for the DNA variant pages, not yet for the RNA pages. The RNA pages therefore contain incorrect examples. I suggest we chahnge these all. So correct would be something like r.456_465TA[4]466_489CAG[9]490_499GCCAG[3].

[...]

Issue #113, opened 2023-12-01
RNA repeat page lists incorrect (?) example


ifokkema:
On the [DNA repeat page](https://hgvs-nomenclature.org/en/latest/recommendations/DNA/repeated/) it is stated that coding transcripts can only have repeat units with a length divisible by three;

> exception: using a coding DNA reference sequence ("c." description) a Repeated sequence variant description can be used only for repeat units with a length which is a multiple of 3

This would then logically also apply to a coding RNA reference sequence. However, the [RNA repeat page](https://hgvs-nomenclature.org/en/latest/recommendations/RNA/repeated/) lists as the first example `r.-124\_-123[14]` and the alternative `r.-124ug[14]`. This can't be correct then; the negative position indicates this is a coding RNA reference sequence, so "ug" can't be a valid repeat sequence.

jfjlaros:
I guess the intent behind this exception is to prevent certain types of repeats in coding DNA, not in `c.` descriptions as a whole. With this interpretation, the `r.` description in the 5' UTR would be admissible.

By the way, we are working on a new proposal for repeats in which we propose to drop this exception.

ifokkema:
@jtdendunnen This hasn't passed community consultation and/or the HVNC committee yet, right? According to the current docs, a community is being prepared. I assume that's what Jeroen and you are talking about. In that case, I guess I'll leave this as is until the new guidelines are out. It makes little sense to "correct" something and have to change it later again.

PR #115, opened 2023-12-01
Fix various issues on the RNA pages

ifokkema:
Fix various issues on the RNA pages.
- Mostly variants not surrounded by `<code>` elements or backticks;
- Various typos;
- Restored one variant description that got edited in a previous conversion step;
- A few incorrect variant descriptions have been highlighted as incorrect examples, and issues have been created (#113, #114) to have them fixed later;
- Don't nest `<code>` elements, as the font size is affected;
- Restored newlines based on the original pages - they got replaced by colons in the new code, now inserted `<br>` tags;
  - Markdown can insert a break by appending two spaces to the last line, but the linter doesn't like that;
  - This is now only applied to the RNA pages. We'll later need to decide on what design/structure to use to unify all pages.


reece (commented):
@ifokkema Thanks for all of the corrections! I agree with all of them.

When we make stylistic changes, my preference would be to focus a PR on that change everywhere. As you point out, changing just the RNA pages means that we also have to attend to the other pages. (In fact, this is why I ended up unifying all of the work while you were away -- it became too difficult to track similar changes in three different branches.)

So, my preference would be for you to create a new issue and branch to use a consistent style for the list with newlines, and to address all of the pages at the same time. Are you open to that?

ifokkema:
Closing this PR in favor of creating a new branch that handles these style fixes in all of the website.

PR #119, opened 2023-12-05
Fix lots of typos

ifokkema:
Fix lots of typos, including misplaced backticks and backslashes used for escaping underscores in variant descriptions that were already surrounded by a `code` element.


reece (approved):
lgtm!

PR #121, opened 2023-12-27
Fixed typo as indicated in Google groups.

ifokkema:
See: https://groups.google.com/g/hgvs-nomenclature/c/UMFUQL4MSrQ/m/pKIvBSB0DAAJ


reece (approved):
Looks good to me. I also checked that the specified coordinates are at intron/exon boundaries.

PR #125, opened 2024-01-08
124 create versions page and release notes for 21.0.0

reece:
Adds new /versions/ pages with a description of the use of versions, release notes for 21.0.0, and a page for historical versions.

Preview at https://hgvs-nomenclature.org/124-create-versions-or-releases-section/versions/


reece:
@ahwagner @jfjlaros @jtdendunnen @ifokkema I would appreciate your comments (or actual edits of you have something major to add) on this pull request.

docs/versions/21.0.md
```diff
@@ -0,0 +1,25 @@
...
+- **Chose a stable URL** To foster stable URLs and better serve the community, HGVS Nomenclature now has a dedicated domain name, [hgvs-nomenclature.org](hgvs-nomenclature.org). The previous pages are now archived at [archive.hgvs-nomenclature.org](https://archive.hgvs-nomenclature.org/). The old site, [varnomen.hgvs.org](https://varnomen.hgvs.org/) forwards users to the corresponding page on the new site.
+- **Modernized our documentation system** Pages from the old documentation were migrated to a system that should make it easier to maintain HGVS Nomenclature and support multiple device size (phones, tablets, laptops). The intention during migration was to retain the intention of the original text, while modernizing the management of HGVS Nomenclature and cleaning up the presentation.
+- **Switched versioning schemes** HGVS Nomenclature will now use Semantic Versioning, which should make it easier for users to track new features in HGVS Nomenclature. Beginning with this release, historical versions will also be available adjacent to newer versions. See [Versions](..) for details.
```
ifokkema, line 15: Should we add instructions how to get to those historical versions?

docs/versions/21.0.md
```diff
@@ -0,0 +1,25 @@
...
+- **Joined HUGO** HGVS Nomenclature is overseen by the [Human Variant Nomenclature Committee](/hvnc/), a subcommittee of the [Human Genome Organization (HUGO)](https://www.hugo-international.org/).
+- **Modernized mechanism to contact the HVNC** We have introduced two new mechanisms to contact the HVNC: there is now an email list to discuss HGVS nomenclature and we have formalized the process by which may submit ideas. See [Contact Us](/#contact-us) for details. The Facebook group and older email address (which was read only by Johan) are now obsolete.
```
ifokkema, line 17: I would suggest changing "Johan" to "the committee's chair".

docs/versions/21.0.md
```diff
@@ -0,0 +1,25 @@
...
+
+### New features
+
+While most of the changes to HGVS Nomenclature were operational, we did introduce a few new features.
+
+- **Introduced formal grammar for HGVS Nomenclature syntax** HGVS Nomenclature previously used human-readable descriptions of the syntax of variants. These descriptions were not usable by computers and sometimes conflicted. We have started to introduce the use of formal grammars to represent variant syntax. See [Grammar Elements](../../recommendations/grammar/) for details.
```
ifokkema, line 23: I think the page linked to needs a "work in progress" label or so. It currently does not contain any variant syntax grammar.

ifokkema (commented):
Great work! I just have some minor suggestions/comments.

docs/versions/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+The HGVS Nomenclature uses concepts from [Semantic Versioning](https://semver.org/) to version releases. Each release version will consist of 3 numbers separated by dots in the format x.y.z with the following meanings:
+
+- The "patch" version (z) is reserved for changes that fix or clarify intentions. This is akin to a bug fix.
+- The "minor" version (y) will be used for new features that are consistent with prior versions but should not break them.  For example, adoption of a community proposal that added a new feature would trigger an increase in the minor version.
+- The "major" version (x) will be used when changes are incompatible with existing conventions, and perhaps for other changes that are deemed to be significant. Major changes should be rare. Wherever possible, the HVNC will strive to ensure that backward incompatibilities result in obvious failures (e.g., falure to parse at all) rather than subtle errors that are difficult to catch.
```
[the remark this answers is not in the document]
reece: What I'm trying to say is that backwards incompatibilities should be rare, and that when they are necessary, we will try to make the failure mode of the incompatibility obvious rather than silent.

For example, suppose we decided to use zero-based numbering. It would be awful to just say that in version 22 that c is zero based because lots of things might parse (and even be semantically plausible). Instead, we might introduce a new t type that is zero based and deprecate c.

So, to be clear, I'm not talking about specific software, but instead our expectation about consequences of changes.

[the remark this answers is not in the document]
reece: Good catch. Changed.

docs/versions/21.0.md
reece: Great idea.

docs/versions/21.0.md
reece: Good point. I'll do that and include in this PR.

docs/versions/21.0.md
```diff
@@ -0,0 +1,25 @@
+# 21.0 Releases
+
+!!! note "We need your help"
+
+    Version 21.0.0 is the first significant update in HGVS Nomenclature in almost four years.
+    While we have resolved many errors during this work, it is possible that new errors were introduced. If you find an issue, please [contact us](/#contact-us).
+
+## 21.0.0 (ETA 2024-01-08)
+
+### Major Changes
```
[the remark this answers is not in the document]
reece: I agree. I experimented with descriptive lists and with `####` headings, but like the current format best. I did update the titles and added periods. Better?

docs/versions/21.0.md
reece: I'm hoping that it's obvious on the interface. Do you find it easy to use? My inclination is to wait to see whether people ask how they get older versions.

reece:
@ifokkema Thanks for the comments. Would you please recheck whether I've addressed them?

docs/versions/21.0.md
```diff
@@ -0,0 +1,25 @@
...
+- **Modernized mechanism to contact the HVNC** We have introduced two new mechanisms to contact the HVNC: there is now an email list to discuss HGVS nomenclature and we have formalized the process by which may submit ideas. See [Contact Us](/#contact-us) for details. The Facebook group and older email address (which was read only by Johan) are now obsolete.
+
+### New features
+
+While most of the changes to HGVS Nomenclature were operational, we did introduce a few new features.
+
+- **Introduced formal grammar for HGVS Nomenclature syntax** HGVS Nomenclature previously used human-readable descriptions of the syntax of variants. These descriptions were not usable by computers and sometimes conflicted. We have started to introduce the use of formal grammars to represent variant syntax. See [Grammar Elements](../../recommendations/grammar/) for details.
+- **Summarized all syntax in a single table** As a by-product of migrating to a formal grammar, we have also generated a summary of all HGVS syntax on a single page by variant type. See [HGVS Syntax Summary](/recommendations/summary/).
+- **Added a list of software that manipulates HGVS expressions** See [Software](/software).
```
ahwagner, line 25: The gene fusion work was a big new feature, adding here. There are probably other syntax-related features we will want to include in this list.

```suggestion
- **Added a list of software that manipulates HGVS expressions** See [Software](/software).
- **Updated recommendations for gene fusion syntax** We have aligned definitions of terms used in the characterization of gene fusions and DNA rearrangements with the VICC Gene Fusion Specification, the HUGO Gene Nomenclature Committee, and the ISCN Nomenclature Committee. This has resulted in the revision of the term _fusion transcript_ to _adjoined transcript_ in relevant sections of the nomenclature and glossary. We have also updated our recommendations for users to consult the [VICC Gene Fusion Specification nomenclature](https://fusions.cancervariants.org/en/latest/nomenclature.html) for exon-based representations of adjoined transcripts.
```

ahwagner (changes requested):
Overall this is great. A few requests for revisions in this review, most minor.

docs/versions/21.0.md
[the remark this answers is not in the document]
reece: @ifokkema Please resolve if you're happy (enough) with the changes made here.

docs/versions/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+    Read the [release notes for the current version](21.0).
+
+
+The primary mission of HGVS Nomenclature is to facilitate reliable communication of sequence variants, which requires that the HGVS Nomenclature is stable. Nonetheless, modifications will be required from time to time in order to address new scientific needs, resolve inconsistencies, or clarify conventions.  A key goal of the [HVNC](/hvnc/) is to manage such changes in a way that balances the needs of progress and reliable data sharing, and to communicate the changes to the community clearly.
+
+The HGVS Nomenclature uses concepts from [Semantic Versioning](https://semver.org/) to version releases. Each release version will consist of 3 numbers separated by dots in the format x.y.z with the following meanings:
+
+- The "patch" version (z) is reserved for changes that fix or clarify intentions. This is akin to a bug fix.
+- The "minor" version (y) will be used for new features that are consistent with prior versions but should not break them.  For example, adoption of a community proposal that added a new feature would trigger an increase in the minor version.
```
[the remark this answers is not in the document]
reece: I use "syntax" and "expression" to mean different things.

In my intended use, syntax is the rule (e.g., EBNF-like) for a specific expression, like a DNA SNV, and an HGVS expression is a specific variant or instance that is consistent with some syntax (=syntactic rule). I've tried to use them consistently in this way, but can't be sure that I always did so.

(By EBNF-like, I merely mean that we can't put it into an EBNF-based parser because we haven't defined all of the productions.)

docs/versions/index.md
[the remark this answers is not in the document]
ifokkema: Clear, thanks!
How about: "(...) ensure that backward incompatible changes result in obvious differences between variant descriptions based on the previous and current version of the nomenclature, rather than subtle differences that are difficult to catch."

docs/versions/21.0.md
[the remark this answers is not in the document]
ifokkema: Excellent! Periods solve the issue, indeed.

docs/versions/21.0.md
ifokkema: I had to look for it, but I'm absolutely fine waiting until people ask.

overrides/partials/version.html
```diff
@@ -4,7 +4,7 @@
 {% set repo_url = config.repo_url + "/releases/tag/" + git.tag %}
 {% endif %}
 
-<a href="{{ repo_url }}">
+<a href="/versions">
```
ifokkema, line 7: This link is broken. That's related to the remark in the `TODO`. A relative link also won't work because that would be different from each page. I was thinking that maybe introducing a `base` element in the HTML would make sense, but then we'd have to adapt all other links in the site, too. I'm not sure what to do here.

overrides/main.html
```diff
@@ -1,23 +1,13 @@
...
-{% block extrahead %}
-
-<!-- Extra style sheets (can't be set in mkdocs.yml due to content hash) -->
-<link rel="stylesheet" href="{{ 'assets/stylesheets/custom.css' | url }}" />
-{% endblock %}
-
 <!-- Announcement bar -->
 {% block announce %}
-<span style="font-size: larger">We've moved! You're looking at the new HGVS Nomenclature pages.</span>
+<span style="font-size: larger">We've moved! You're looking at the new HGVS Nomenclature pages. <a href="/versions/21.0/">What's Changed?</a></span>
```
ifokkema, line 5: This link is also broken, for the same reason. Since it also won't work on the stable site, perhaps we should remove it? Or we should fix the issue properly.

ifokkema:
Hi @reece, we're getting there! I added some commits, mostly to fix links. Please do double-check [this commit](https://github.com/HGVSnomenclature/hgvs-nomenclature/pull/125/commits/d4561b82dc6666a513a624c2fa4db13eda542d89) as I removed some stuff that was causing 404 errors and I'm 99% sure it wasn't doing anything, but you should be able to tell for certain.

docs/versions/index.md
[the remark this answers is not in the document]
reece: I think I see what you're getting at regarding the implication of some sort of test. How about this?

Wherever possible, the HVNC will strive to ensure that incompatible differences between versions are likely to (or can be made to) result in obvious failures when implemented in software (e.g., failure to parse at all) rather than subtle errors that are difficult to catch or might be overlooked.

overrides/partials/version.html
reece: I agree that this is a problem. I am stuck with how to resolve it. Perhaps we should just remove it for now?

docs/versions/index.md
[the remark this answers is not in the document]
ifokkema: Looks great!

overrides/partials/version.html
ifokkema: That would probably be the best idea, reverting to the link to Github. We can then later see how to solve issues like these.

overrides/main.html
reece: So annoying! Having relocatable sites but need the ability to link to documents relative to the root seems nearly incompatible. I'm stuck on this one too. I'll remove this and the versions link.

reece:
> Hi @reece, we're getting there! I added some commits, mostly to fix links. Please do double-check [this commit](https://github.com/HGVSnomenclature/hgvs-nomenclature/pull/125/commits/d4561b82dc6666a513a624c2fa4db13eda542d89) as I removed some stuff that was causing 404 errors and I'm 99% sure it wasn't doing anything, but you should be able to tell for certain.

Yep, you're right. Thanks for resolving this!

docs/versions/index.md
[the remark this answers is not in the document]
ifokkema: I agree that syntax is a better term; I always use "syntax" to explain the rule, e.g., the conversion syntax being removed in favor of the delins syntax. For usages of this syntax, I generally use the term "variant description". Although I am a programmer, thus more familiar with the word syntax, it's currently used throughout the website. If "syntax" is too technical of a term, all pages using it should be reworded.

overrides/main.html
ifokkema: Indeed :thinking:  I usually use the `base` element in my websites to solve this, but that would be too much of a hassle now. We can try to solve it later.

docs/versions/index.md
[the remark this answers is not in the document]
reece: So, @ahwagner and @ifokkema, can we agree to use "syntax" to refer to the rule? 

And, "variant description" seems more transparent and accessible to readers than "HGVS expression", so I'm fine with than. 

In other words, let's do what @ifokkema said. Good?

overrides/partials/version.html
reece: I worked out something that I like better. Let's see what you think, @ifokkema.

I made https://hgvs-nomenclature/versions redirect to https://hgvs-nomenclature/stable/versions on RTD. This means that the absolute URL will always resolve to the most recent stable documentation version. IMO, that's 99% of the goal.

In branch deployments, the correct versions will be be accessible in the nav menu. The only thing that will be broken is that clicking the version tag in the upper right will take the user (a developer, likely) to the stable versions page. This seems like a small annoyance. 

And local development will always give the local version of the versions page.

The redirect is working now (but 404 because the versions page doesn't exist yet). If you agree, I'll leave the URL as /versions/.

overrides/main.html
reece: I think this is the same as previous. I vote to leave as-is. We can't meaningfully test it until we merge this branch.

reece:
@ahwagner @ifokkema : I think all of the issues are resolved as well as we can. I would be grateful for a quick check and approval if you agree. Let's launch!

docs/versions/index.md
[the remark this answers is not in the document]
ahwagner: Yes, I agree. Updating `HGVS expression` -> `(HGVS)? variant description` throughout the site appears to remain open. Making a commit to address.

ahwagner:
I tried merging `main` back in to this branch in prep for merge but it was a larger change that I expected, so reconsidered and forced the branch back to 619b996.

ahwagner (approved):
Beyond the need to bring this branch up-to-date with `main`, I think this PR looks ready to go.

reece:
@ahwagner @ifokkema Thanks for all of the changes and reviews from both of you. Alex approved. Merging!

overrides/partials/version.html
ifokkema: Sure, this will be OK! And now that the site is live, the link works.

overrides/main.html
ifokkema: This link now also works!

PR #130, opened 2024-01-10
129 - Revise use of "fusion transcript" to "adjoined transcript"

ahwagner:
Addresses #129, disambiguating between gene fusions, rearrangements, and adjoined transcripts

Issue #129, opened 2024-01-10
Fusions - adjoined transcript


ahwagner:
Some revisions from "fusion transcript" to "adjoined transcript" were lost or reverted during the transition to hgvs-nomenclature.org and missed in #22.

The RNA pages need to be updated with this change.

reece:
@ahwagner Is this now closed?

ahwagner:
Yes, this was completed in #131

PR #131, opened 2024-01-10
129 - Revise use of "fusion transcript" to "adjoined transcript" (attempt 2)

ahwagner:
Another pass at #130. Diff appears correct this time.

Addresses #129, disambiguating between gene fusions, rearrangements, and adjoined transcripts.


ifokkema (approved):
Looks good to me! I ran into some typos that I fixed, and I also took the liberty of styling the variant descriptions on the SVD-WG007 page while I was at it.

PR #138, opened 2024-01-19
Update deletion.md

### LINKED ISSUE #137: Review the first example on the Deletion page
The text/HGVS is inconsistent with the sequence example shown.  The deleted NT in the sequence shown is an A but the NT in the text and in the HGVS expression is a T:

- one nucleotide

-     NC_000023.11:g.33344591del: a deletion of the **T** at position g.33344591 in the sequence TGTG**A**TTCT to TGTG**_**TTCT: NOTE: the recommendation is not to describe the variant as NC_000023.11:g.33344591del**T**, 

If the deleted NT in the sequence is changed to T, it should be right-shifted.
https://hgvs-nomenclature.org/stable/recommendations/DNA/deletion/#examples

jfjlaros:
Fixes #137.


reece (approved):
lgtm. I double checked the nucleotide. Thanks for fixing @jfjlaros !

PR #140, opened 2024-02-05
add group email address to top Contact Us section

reece:
In response to suggestion from
https://groups.google.com/g/hgvs-nomenclature/c/toGJA-HsPvE/m/SrDheuHoAAAJ


ifokkema:
Should we let people know that the answer may actually only be found in the Google group? Because, when replying online, the message is only sent to the group and not sent to the poster. It's unclear to me whether the poster is notified of a reply to their question. We've had one person reminding us of a question that was already answered.

reece:
@ifokkema I added a comment about sending messages and reading replies. Does this address your request?

PR #145, opened 2024-02-05
Remove "indel" except for where it's made clear it shouldn't be used.

### LINKED ISSUE #142: Remove the term "indel" where needed.
The nomenclature states that "indel" shouldn't be used;

> The term "indel" is not used in HGVS nomenclature (see [Glossary](https://hgvs-nomenclature.org/stable/background/glossary/)). The term is confusing, having different meanings in different disciplines.

([source](https://hgvs-nomenclature.org/stable/recommendations/RNA/delins))

However, "indel" is used quite a few times in the docs. Replace those mentions with "delins".
Work will be done as part of #134 to prevent unnecessary conflicts with that branch.

ifokkema:
Replaced all mentions with "delins" except for in the hgvs package description; there, I changed the sentence to:

> A key feature of the hgvs package is the ability to correctly project variants in the presence of genome-transcript alignment **deletion/insertions** (...)

@Reece, please change it if you prefer other wording.

Merging into `134-fix-all-remaining-style-issues` to prevent lots of conflicts later, as that branch is changing all the lines that I'm editing now as well.

Closes #142.

### DIFF
--- a/docs/background/simple.md
+++ b/docs/background/simple.md
@@ -80,7 +80,7 @@ All variants given are in the DMD gene and reported in relation to coding DNA re
     - c.4375_4385dup: the nucleotides from position c.4375 to c.4385 (CGATTATTCCA) are present twice (duplicated). Often reported as c.4375_4385dupCGATTATTCCA or c.4385_4386insCGATTATTCCA (not a correct HGVS description).
 - insertion: one or more letters in the DNA code are new (inserted). An insertion is indicated using **"ins"**.
     - c.4375_4376insACCT: the new sequence "ACCT" was found inserted between positions c.4375 and c.4376.
-- deletion/insertion (indel): one or more letters in the DNA code are missing and replaced by several new letters. A deletion/insertion is indicated using **"delins"**.
+- deletion/insertion (delins): one or more letters in the DNA code are missing and replaced by several new letters. A deletion/insertion is indicated using **"delins"**.
     - c.4375_4376delinsAGTT: the nucleotides from position c.4375 to c.4376 (CG) are missing (deleted) and replaced by the new sequence "AGTT". Also reported as c.4375_4376delCGinsAGTT.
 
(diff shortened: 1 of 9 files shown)


jfjlaros (approved):
This looks good to me.

In general, do we always use the term "deletion/insertion" in running text?

reece (approved):
LGTM too. Thanks for the cleanups and for asking about the hgvs block.

ifokkema:
@jfjlaros 
> In general, do we always use the term "deletion/insertion" in running text?

The site uses a combination of "deletion/insertion" and "deletion-insertion". Looking at the stats, the latter is actually more common. I guess it would be good if we would standardize this later as well.

@reece 
> LGTM too. Thanks for the cleanups and for asking about the hgvs block.

No worries! FYI: The general cleanup/styling is moving along slowly, but steadily.

jfjlaros:
@ifokkema 
> I guess it would be good if we would standardize this later as well.

Agreed.

I also see we use the term `delins` in [running text](https://github.com/HGVSnomenclature/hgvs-nomenclature/blob/9a99e46cbf021826dd4ba07837f181ae6248fa5f/docs/recommendations/RNA/delins.md?plain=1#L17) as well. I think it would be better to use common names (deletion, insertion, etc.) instead of operator names (`del`, `ins`, etc.) in running text.

ifokkema:
@jfjlaros 

> I also see we use the term `delins` in [running text](https://github.com/HGVSnomenclature/hgvs-nomenclature/blob/9a99e46cbf021826dd4ba07837f181ae6248fa5f/docs/recommendations/RNA/delins.md?plain=1#L17) as well. I think it would be better to use common names (deletion, insertion, etc.) instead of operator names (`del`, `ins`, etc.) in running text.

I agree. Currently, however, I'm working on a large update of every single page in the docs, and at least half of the lines (and often, more) are affected. So it would be great if we could do this only after I complete that work. Otherwise, I'll have to deal with a lot of conflicts. I'm also making a list of things that we should fix after I'm done.

jfjlaros:
Sure, we can put it on our backlog.

PR #149, opened 2024-02-19
fix: DNA inversion syntax typo

markgene:
"dup" is used for DNA inversion in the current syntax.


reece (approved):
Good catch. Thanks @markgene !

PR #151, opened 2024-02-22
Update complex.md

jtdendunnen:
by definition "::" is used to indicate the fusion molecule, it is not necessary what is deleted, this is obvious from the "::".


ifokkema (commented):
I find the modification confusing; in the original description, `pter_(12200001_14700000)del` seemed to indicate that `pter` for sure is deleted; now, it indicates the opposite. Also, `::` is also in use for adjoined transcripts, and there, each end should have a start and end position, not just the position where the breakpoint is. I find this an unnecessary conflict between how different joins of molecules are described.

Also, further up the same page, an example is given for `NC_000022.11:g.pter_(12200001_14700000)del::(37600001_410000000)_qterdel`. Those are the same positions, and the examples are clearly related since the ISCN description is nearly identical; `r(22)(p11.1q13.1)` vs `+r(22)(p11.1q13.1)`. Should the other example also be updated?

PR #154, opened 2024-03-06
Typo in Protein Duplication

mbiokyle29:
While reviewing the page for protein duplication, I think I noticed a typo, where the range syntax specifies you should use the `delins` slug instead of the `dup` slug and should include the `sequence`, even though the example has `dup` with no sequence. I just updated the `syntax.yaml` file to address this (not sure if that is the right solution, this is my first contribution).

The issue, as seen on the main site:
Screen Shot 2024-03-06 at 9 38 32 AM

Addressed running locally on my branch:
Screen Shot 2024-03-06 at 9 58 59 AM

### DIFF
--- a/docs/syntax.yaml
+++ b/docs/syntax.yaml
@@ -69,7 +69,7 @@ aa:
         examples:
           - NP_003997.2:p.Val7dup
       - name: position range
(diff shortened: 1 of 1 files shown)


ifokkema (approved):
Good catch! Thanks!

PR #159, opened 2024-03-19
Fix description for supernumerary chromosome

### LINKED ISSUE #153: What is the correct description for a supernumerary chromosome?
The HGVS nomenclature explains two ways to describe supernumerary chromosomes.

With square brackets:
> the description of the supernumerary molecule is given using "[ ]sup".

([source](https://hgvs-nomenclature.org/stable/recommendations/DNA/complex/))

or, without:
> `NC_000023.11:g.pter_qtersup`

([source](https://hgvs-nomenclature.org/stable/recommendations/DNA/duplication/))

I have just received data where the description `NC_000023.10:g.[pter_qter]sup` is given, and I can't tell whether that's correct or not.

ifokkema:
### Fix the description for a supernumerary chromosome
- Fix the suggestion of using `[ ]sup` which should be `sup`. Removed the whole line since it added nothing to the already existing paragraph.
- Fix the naming as well; supernumary should be supernumerary.

Closes #153.


jfjlaros (approved):
I agree with this.

PR #162, opened 2024-03-21
Fixed an example on the DNA allele page.

### LINKED ISSUE #58: Correct an answer given on the DNA allele page.
While correcting some mistakes in some examples, I ran into [this one](https://github.com/HGVSnomenclature/hgvs-nomenclature/blame/3f61cfd6f2df135da4212519f546334de93d42c8/docs/recommendations/DNA/alleles.md#L76):

> > I have a patient with hearing loss and variants in the GJB2 (c.35delG) and GJB6 (c.689_690insT) genes, how should I describe this? (Nancy Carson, Ottawa, Canada)

> The recommendation is to use the format GJB2:c.[35delG] GJB6:c.[689_690insT]. This uses standard HGVS descriptions and prevents confusion regarding which variant was found in which gene. Note it is essential that you also define the coding DNA reference sequence used. Another format, coping with this directly, is to describe the variants as NM_004004.2:c.[35delG] NM_006783.1:c.[689_690insT], i.e. using the Genbank reference sequences in stead of the HGNC approved Gene Symbol.

This must be very old; it can't seriously be the recommendation... I suggest:
- [x] Fix the question to include that these variants are _in trans_. That's not currently made clear and was only clarified during the discussion on this example.
- [x] Fixing the square brackets from the variant descriptions in this answer. These variants are _in trans_.
- [x] Changing the `NM_004004.2:c.35delG` to `NM_004004.2:c.35del`, as the `G` is redundant.
- [ ] ~~Perhaps changing the `NM_004004.2:c.35del` to either `NM_004004.2(GJB2):c.35del` or `GJB2(NM_004004.2):c.35del`; although I can't find whether this is official syntax, the first one has been used once elsewhere in the documentation, and the latter is in use in LOVD (as far as I remember, approved by Johan).~~ So this may require discussion. There was some talk in a recent HUGO Gene/Disease Specific Database Advisory Council meeting about this format, as a curator requested it.
- [x] Perhaps removing the descriptions without the reference sequences (e.g., `GJB2:c.35del`) to discourage its use. Or, otherwise, moving it to the back of the answer as a shorter alternative with the prerequisite to have defined the transcript reference sequence elsewhere.

ifokkema:
#### Fixed an example on the DNA allele page.
- Clarify that the question is about variants _in trans_.
- Fix the square brackets from the variant descriptions based on the other given example.
- Change `c.35delG` to `c.35del`, as the G is redundant.
- Changed the order of the provided descriptions, promoting the use of the one using reference sequences.

Closes #58.


jfjlaros (commented):
I think we should finish discussion #58 before merging this PR.

reece (approved):

PR #166, opened 2024-04-17
Remove call for members.

ifokkema:
### Remove call for members.
- Remove all references to the Call for Members page. I decided to keep all the references as Markdown comments. That way, we can easily put them back whenever we have a new call.
- Add an expiration notice on the call page in case people reach it. Whether it will be built depends on how `mkdocs` is invoked, and in case we'll still have it online and people reach it through Google or so, they'll know it's expired and how to watch for new calls.

### DIFF
--- a/docs/call-for-members.md
+++ b/docs/call-for-members.md
@@ -1,5 +1,12 @@
 # Call for Members
 
+!!! warning "Expired"
+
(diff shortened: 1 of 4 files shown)


reece (approved):
lgtm

PR #169, opened 2024-04-17
coding/non-coding identity

jtdendunnen:
as discussed in HVNC meeting, there should be no difference in the description of this rule for coding and non-coding reference sequences


ifokkema (approved):

ifokkema:
@jfjlaros I believe this was the sentence that you mentioned, correct?

jfjlaros:
Indeed.

Additionally, instead of using "reference sequence", I would propose to use something like "transcript annotation", e.g.,  "it is **not** allowed to describe variants in nucleotides beyond the boundaries of the annotated transcript."

jfjlaros:
Indeed, it should never be allowed to address a coordinate outside of the reference sequence, regardless of the coordinate system used.

Perhaps we can make this a general remark and specifically mention the transcript boundaries in the "coding" and "noncoding" sections.

ahwagner (commented):
I'm not sure I follow this change; if we are disallowing representation of variants outside of a specified reference (which I think is a good thing), when would we use +/- syntax? I am assuming this is still intended to be used since it is not being removed in this PR.

Is there an associated discussion / issue thread to provide context for what this PR is addressing?

jfjlaros:
> when would we use +/- syntax?

When we use a genomic reference sequence.

For example, description [NG_012337.3(NM_003002.4):c.52+100del](https://mutalyzer.nl/normalizer/NG_012337.3(NM_003002.4):c.52+100del) is valid because position `c.52+100` indexes a `G` at position 5213 in the reference sequence of `NG_012337.3`.
On the other hand, description [NM_003002.4:c.52+100del](https://mutalyzer.nl/normalizer/NM_003002.4:c.52+100del) is invalid since `c.52+100` does not index anything in the reference sequence of `NM_003002.4`.

reece (approved):

docs/background/numbering.md
```diff
@@ -51,7 +51,7 @@ The latter format, `c.IVS2+2T>G` / `c.IVS2-1G>T`, has been retracted and **shoul
 
 - nucleotide numbering is `n.1`, `n.2`, `n.3`, ..., etc., from the first to the last nucleotide of the reference sequence.
 - nucleotides in introns are numbered as for coding DNA reference sequences [(see above)](#DNAc), although proceeded by `n.` (not `c.`).
-- it is **not** allowed to describe variants in nucleotides which are not covered by the transcript, using only a non-coding DNA reference sequence.
+- it is **not** allowed to describe variants in nucleotides beyond the boundaries of a reference sequence, using that reference sequence.
```
ahwagner, line 54: From 10/7/23 committee meeting: `that` should be updated to `the transcript` reference sequence. Should result in:

> it is **not** allowed to describe variants in nucleotides beyond the boundaries of a _transcript_ reference sequence, using _that transcript_ reference sequence

ahwagner (commented):
In addition to the specific requested change, it was noted that this edited text continues to conflict with other parts of the website, which should now be updated to match this.

docs/background/numbering.md
ifokkema: Fixed in ab8d640.

ifokkema:
> In addition to the specific requested change, it was noted that this edited text continues to conflict with other parts of the website, which should now be updated to match this.

Fixed in https://github.com/HGVSnomenclature/hgvs-nomenclature/commit/b167111. I will open a PR.

Issue #168, opened 2024-04-17
Create a new release, to update the "stable" tag.


ifokkema:
During our last HVNC meeting, we couldn't easily locate the updated figure on the numbering page. This is due to the site's `stable` tag linking to the latest version and not to the `main` branch. That is fine, but that means we need more frequent updates if we want our updates to reach the public sooner. It would be good to discuss when to release and which of us can do so. I'm fine with having a go at it, but I don't want to do so without discussing it first with @reece and @ahwagner.

reece:
@ifokkema Please see docs that I added in a PR. Want to try to release 21.0.3?

ifokkema:
@reece That worked great! Thanks for the guidance!

reece:
Terrific! Thanks for taking it on @ifokkema !

PR #172, opened 2024-07-07
remove commented call for members 

reece:
#166 commented out the call for new members. However, Read The Docs appear to not honor the comment notation and the call for new members remains (see https://hgvs-nomenclature.org/stable/). This PR removes the call entirely. (But it was a good idea, @ifokkema!).

### DIFF
--- a/docs/hvnc.md
+++ b/docs/hvnc.md
@@ -1,10 +1,5 @@
 # HGVS Variant Nomenclature Committee (HVNC)
 
-[//]: # (!!! note inline end "Join Us")
-[//]: # ()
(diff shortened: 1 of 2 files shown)


ifokkema:
Note that the /stable branch hasn't been updated in a long time. So the call remains simply because it's currently running [this version](https://github.com/HGVSnomenclature/hgvs-nomenclature/releases/tag/21.0.2).
See also https://github.com/HGVSnomenclature/hgvs-nomenclature/issues/168.
No worries, it confused us as well :sweat_smile:

reece:
Doh! You're right, of course. I knew about this behavior, but I didn't realize (or check) that we hadn't released a new version. 

I'll reply in #168 about the rest.

PR #176, opened 2024-08-12
Automate release notes

reece:
This PR streamlines the release flow by pulling the release notes from GitHub directly into the generated docs at build time. These notes obviate the release notes that were maintained manually in docs/versions/21.0.md. The release process docs (dev-docs/release-new-version.md) were updated appropriately.

## Advantages
- Simplicity: Release is as simple as using the github release process and the site will incorporate the notes automatically
- Consistency between github release notes and the versions page

## Disadvantages
- No mechanism to differentiate github release notes and site release notes
- Small possibility of build failures due to API failure (including API rate limits)
- Pages are regenerated for all versions on every build. (In practice, this is «1 sec) 

Most of these disadvantages can be overcome by creating and committing version page as before, and then making the page for only the most recent major.minor. 

Comparing 
https://hgvs-nomenclature.org/stable/versions/21.0/ and 
https://hgvs-nomenclature--176.org.readthedocs.build/176/versions/21.0/
shows that these pages are nearly identical with the exception of the admonition.


ifokkema (approved):
This is a great idea, and it works really well!

I managed to get the output exactly the way it was, and I fixed the release notes for 21.0.0, which were lost. If you don't like the way I did that, feel free to implement it differently. Otherwise, it's good to go!

reece:
Nice changes, @ifokkema .

I removed your inline 21.0.0 release notes and created a new 21.0.0. release in github. The output is essentially identical, but it means that users will also now see a 21.0.0 release in github and that the release notes work the same as the others. Thanks for catching that omission!

Since this is essentially a refinement on code that you already reviewed, I'm going to merge as-is.

ifokkema:
> I removed your inline 21.0.0 release notes and created a new 21.0.0. release in github.

Makes sense! I didn't realize it would nicely sort the way it does; I thought a new release would just get in the way. This is much better, indeed! Thanks!

PR #181, opened 2024-09-25
Update HVNC members

ifokkema:
### Update HVNC members
- Remove members who recently left the HVNC; add them as past members.
- Add the new members to the table.

@reece @ahwagner @jfjlaros I messed up and pushed to `main` because I forgot to create a branch. Not sure why it let me push to `main`, though. It did give me a warning, but it continued anyway. I now put everything in a new branch, but I can't reset `main` as I can't force-push to it. So these changes are already in `main`, but by accident. I could commit a fix to `main`, removing the changes, or we can leave it, in case this PR is good to go anyway.

### DIFF
--- a/docs/hvnc.md
+++ b/docs/hvnc.md
@@ -18,15 +18,19 @@ The mission of the HVNC is to define the standard for the unambiguous descriptio
 
 The HVNC started in 2013, initiated by Global Variome (GV) / the Human Variome Project (HVP).
 
-| Name                         | Institution                                  |    Term     |
-| :----------------------------| :-----------------------------------         | :---------: |
-| Johan den Dunnen (chair)     | Leiden University Medical Center, Nederland  | 2013 - .... |
-| Marina DiStefano             | Broad Clinical Labs, United States           | 2022 - 2025 |
-| Ivo Fokkema                  | Leiden University Medical Center, Nederland  | 2023 - 2026 |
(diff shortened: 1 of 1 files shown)


reece:
Hi @ifokkema . Thanks for trying to clean this up. 

Since this is a small and uncontrovertial change, I'd just leave it as-is and not bother with an after-the-fact PR.

Code owners are permitted to bypass the requirement for a branch and PR. In GitHub, you'd get bright red buttons asking to confirm. There's no such thing with git, so it just pushes with a warning.

The cost of changing that setting is that everything and everyone would require a PR, which seemed pretty cumbersome to me.

reece:
P.S. You could revert commits with `git revert`, which just creates a patch in the opposite direction. That's better than force pushing.

Again, I think we should just let it be, so I'm going to close the PR. @ahwagner @jfjlaros, please reopen if you feel differently.

ahwagner:
Agree. Changes are non-controversial and fine to leave as-is. Arguably even minor changes are appropriate to make as PRs to allow for commentary for a few days, though agree that mandating review / approval for uncontroversial edits like this would be an unhelpful change in repo administration policy.

ifokkema:
Thanks, @reece, for the clarification and guidance on `git revert`. And thanks, @ahwagner, for putting that setting into perspective; it makes total sense. I guess I am less careful because sometimes I use `main` in projects I work on by myself. I should be able to create a git hook for this repo only that prohibits me from committing locally in `main`.

PR #184, opened 2024-10-07
2024-10-07 meeting updates

ifokkema:
### 2024-10-07 meeting updates
- Update the numbering page as discussed during today's meeting.
- Fixed error in an answer on the DNA substitution page. This was discussed in today's meeting. Using c. positions beyond the boundaries of the transcript reference sequence is not allowed when using only that transcript reference sequence, even though it's mentioned as a possibility in this answer. That is an error. However, today, we voted to allow using the c. coordinate system for intergenic positions as long as a genomic context was added. The answer has been corrected, and this new decision has been clarified.

### DIFF
--- a/docs/recommendations/DNA/substitution.md
+++ b/docs/recommendations/DNA/substitution.md
@@ -74,7 +74,7 @@ bin/pull-syntax -f docs/syntax.yaml dna.sub
     Describing the variant in relation to a coding DNA reference sequence is only possible when the nucleotide is included in this transcript reference sequence, or when a genomic sequence context is added.
     E.g., <code class="invalid">NM_004006.1:c.-128354C>T</code> or <code class="invalid">NM_000109.3:c.-401C>T</code> are invalid, because the positions `c.-128354` and `c.-401` are not included in the transcript reference sequences used.
     However, when adding in the context of a genomic reference sequence, this variant can be described as `NC_000023.10(NM_004006.1):c.-128354C>T` or `NC_000023.10(NM_000109.3):c.-401C>T`.
-    The variant can also be described using a genomic reference sequence containing the promoter region (for this variant e.g., `L01538.1:g.1407C>T`), but again this is not really informative.
+    The variant can also be described using a genomic reference sequence containing the promoter region (for this variant e.g., `L01538.1:g.1407C>T`).
     Although `NC_000023.10:g.33357783G>A` seems complex, it can be used in a genome browsers helping you to quickly zoom in on the region of interest.
 
 <a id="polymorphism"></a>
--- a/docs/recommendations/DNA/substitution.md
+++ b/docs/recommendations/DNA/substitution.md
@@ -70,8 +70,14 @@ bin/pull-syntax -f docs/syntax.yaml dna.sub
 
 !!! note "How should I describe a variant in the promoter region of a gene?"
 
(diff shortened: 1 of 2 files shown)


docs/recommendations/DNA/substitution.md
```diff
@@ -70,8 +70,14 @@ bin/pull-syntax -f docs/syntax.yaml dna.sub
...
-
+    It is recommended to describe variants in the promoter region of a gene based on a genomic reference sequence, e.g., `NC_000023.10:g.33357783G>A` (chrX, hg19).
+    Describing the variant in relation to a coding DNA reference sequence is only possible when the nucleotide is included in this transcript reference sequence, or when a genomic sequence context is added.
+    E.g., <code class="invalid">NM_004006.1:c.-128354C>T</code> or <code class="invalid">NM_000109.3:c.-401C>T</code> are invalid, because the positions `c.-128354` and `c.-401` are not included in the transcript reference sequences used.
+    However, when adding in the context of a genomic reference sequence, this variant can be described as `NC_000023.10(NM_004006.1):c.-128354C>T` or `NC_000023.10(NM_000109.3):c.-401C>T`.
+    The variant can also be described using a genomic reference sequence containing the promoter region (for this variant e.g., `L01538.1:g.1407C>T`), but again this is not really informative.
```
ahwagner, line 77: Suggest removing the exposition here.

```suggestion
    The variant can also be described using a genomic reference sequence containing the promoter region (for this variant e.g., `L01538.1:g.1407C>T`).
```

ahwagner (changes requested):
Mostly LGTM, one small edit requested.

docs/recommendations/DNA/substitution.md
ifokkema: Yeah, that makes sense; I'll add your correction.

docs/recommendations/DNA/substitution.md
```diff
@@ -70,8 +70,14 @@ bin/pull-syntax -f docs/syntax.yaml dna.sub
...
 
-    It is recommended to describe variants in the promoter region of a gene based on a genomic reference sequence, e.g., `NC_000023.10:g.33357783G>A` (chrX, hg19). Describing the variant in relation to a coding DNA reference sequence (for this variant `NM_004006.1:c.-128354C>T` or `NM_000109.3:c.-401C>T` is possible but not really very informative; you do not know how long the 5'UTR is. The variant can also be described using a genomic reference sequence containing the promoter region (for this variant e.g., `L01538.1:g.1407C>T`), but again this is not really informative. Although `NC_000023.10:g.33357783G>A` seems complex, it can be used in a genome browsers helping you to quickly zoom in on the region of interest.<a id="polymorphism"></a>
-
+    It is recommended to describe variants in the promoter region of a gene based on a genomic reference sequence, e.g., `NC_000023.10:g.33357783G>A` (chrX, hg19).
+    Describing the variant in relation to a coding DNA reference sequence is only possible when the nucleotide is included in this transcript reference sequence, or when a genomic sequence context is added.
```
jfjlaros, line 74: I am a bit late to the party, but we cannot say that a genomic sequence context was added. See discussion #182.

jfjlaros (commented):
I propose to rephrase or revert part of this merge until discussion #182 is settled.

docs/recommendations/DNA/substitution.md
ifokkema: > (...) but we cannot say that a genomic sequence context was added.

Why not? Both interpretations of NC(NM) add a genomic context.

jfjlaros: Only one does, the other one only operates within a genomic context (nothing is "added").

ifokkema: The context is added to the variant by the reference sequence... but if somehow that word doesn't work for you, we can use "used"?

jfjlaros: I would suggest: "[...] when a genomic sequence is used."

reece:
@jfjlaros : Since this has already been merged, would you please submit a new PR with the reversion and/or replacement text that you'd like to see?

jfjlaros:
Can do. Please see #203.

PR #188, opened 2024-10-18
Update contrib guide

ahwagner:
Adds text drafted by @marinadistefano and reviewed by HVNC.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -36,8 +36,8 @@ expedited response in relation to a clinical case, please indicate this in both
 the subject line and body of the email inquiry. We cannot guarantee that even
 clinical inquiries will be addressed outside of business hours or on weekends.
 
-You may also submit questions to our [GitHub repository Q&A board]
-(https://github.com/HGVSnomenclature/hgvs-nomenclature/discussions/new?category=q-a).
+You may also submit questions to our
+[GitHub repository Q&A board](<https://github.com/HGVSnomenclature/hgvs-nomenclature/discussions/new?category=q-a>).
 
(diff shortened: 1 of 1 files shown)


marinadistefano:
This looks good to me! Thanks @ahwagner

reece (approved):

ifokkema:
This also adds a reference to a method to ask us a question on GitHub:
> You may also submit questions to our
[GitHub repository Q&A board](<https://github.com/HGVSnomenclature/hgvs-nomenclature/discussions/new?category=q-a>).

Is it a good idea to offer multiple methods for this? Is one preferred over the other? Does the deadline that we set for ourselves also count for the GitHub method of asking us a question?

marinadistefano:
Yeah, I do have concerns about missing something that it on our GitHub Q&A board instead of via email.

reece:
Ah. FWIW, I read the PR as this already being approved by the HVNC so didn't comment.

If it's still up for discussion, then 1) I agree that we should drop the Q&A board, 2) I think the text about the mailing list usage, turn around time, etc. is far too long. (Preview at https://hgvs-nomenclature.org/main/).

I recommend:
- merging the first two paragraphs ("Join the..." and "Messages sent to...").
- dropping the entire "When members of the community..." paragraph
- dropping the Q&A board paragraph

Also, in a separate PR, I proposed to rename this section and add a new announcement-only group. See https://hgvs-nomenclature--191.org.readthedocs.build/191/.

reece:
Please continue the conversation on #194

PR #190, opened 2024-11-03
Fix mkdocs issues

reece:
This PR addresses several small mkdocs administrative shortcomings:

- logging wasn't working in hooks.py
- mkdocs-extras was no longer preventing `make serve` from endless rebuilds (files in docs/versions/ would get rebuilt, triggering another rebuild pass)
- ...so, versions files are now updated only if changed (and can also be built with `make update-release-notes`).
- requirements weren't pinned at all (and are now pinned to major.minor)


reece:
@ahwagner Since you appear to be around... I'd appreciate a review on this. Just tending to some weeds in our mkdocs setup.

ifokkema:
I might be wrong, but I think the changes in this PR would require us also to update the [release docs](https://github.com/HGVSnomenclature/hgvs-nomenclature/blob/main/dev-docs/release-new-version.md)?

reece:
> I might be wrong, but I think the changes in this PR would require us also to update the [release docs](https://github.com/HGVSnomenclature/hgvs-nomenclature/blob/main/dev-docs/release-new-version.md)?

I don't think so, but I might have created too many ways for version pages to be created, with subtle differences in outcomes, and therefore created confusion. I think the addition of the Makefile target should be considered a parallel and experimental route to generating for debugging purposes, not the intended process.

PR #191, opened 2024-11-03
Add description of new hgvs-nomenclature-announcement group

reece:
A reviewer suggested that we have a second group for announcements only.

I think this is a good idea (and we'd discussed it previously). I created the group but didn't announce it anywhere. This PR adds that group to the top-level page.

If reviewers are in agreement, please merge this PR. 

If there is concern with adding another group, don't merge this PR and we'll discuss.


ahwagner (approved):
I agree, but would want at least one other person to approve before merge.

ifokkema:
Merging may depend on #194, as in, we may make the fixes discussed there, in this PR.

Regarding the changes currently in this PR, I'm not against merging, but I do find the loss of the "Contact us" a bit strange if we keep the part where we discuss how fast we'll be replying. That kind of implies that the person is contacting us, and not engaging with the community.

Unrelated to the content, I see the changes in this PR are sometimes not actual changes but a rewrapping of the existing content. Is that caused by settings within an IDE? Perhaps we should all take the same settings (or disable rewrapping of untouched content) to prevent diffs that are larger than needed. It makes me stare at the screen, trying to figure out what changed in that paragraph, only realizing after a while that the only thing changed was whitespace :sweat_smile:

reece:
I agree that whitespace changes are annoying when mixed with substantive changes. I usually avoid making such changes, but I think they're inevitable when moving paragraphs around (as happened here). If there are unnecessary whitespace changes, I apologize.

How would you feel if we merged this into main, then tackled #194 as a holistic rethinking of the entire section?

docs/index.md
```diff
@@ -10,39 +10,47 @@ The HGVS Nomenclature is administered by the [HGVS Variant Nomenclature Committe
...
+archive**.
 
-When members of the community reach out to the HGVS nomenclature committee with
-questions via the google group email, we aim to respond within 2 business days
-of email receipt. If you haven’t received a response within four business days,
-please post a reminder under your question in the google group. For a more
-expedited response in relation to a clinical case, please indicate this in both
-the subject line and body of the email inquiry. We cannot guarantee that even
-clinical inquiries will be addressed outside of business hours or on weekends.
+We aim to respond within two business days of email receipt. If you haven't
```
ahwagner, line 39: We should not be overwriting this text added in #188 without committee discussion.

ahwagner (changes requested):
New revisions have been added to this PR since my last review that undo HVNC-vetted text from #188. I recommend reverting those changes before merge.

reece:
I reverted to main, then added just the blurb about the announcement group. We can pick up rewording this section in #194 (PR: #200). Please re-review.

ahwagner (approved):
LGTM (again)!

ifokkema (commented):
LGTM!
(edit: I saw too late that it was merged already; ignore my useless comment, please :grin: )

Issue #143, opened 2024-02-05
Optimize search feature


ifokkema:
The search feature could use some fixing. I don't know how it's implemented, so this may require discussion or somebody else to step in and fix it.

![Screenshot_2024-02-05_12-13-19](https://github.com/HGVSnomenclature/hgvs-nomenclature/assets/11665018/ef0e2167-76ba-4428-a7f9-c6d0e36a6587)

When searching for indel, I've noticed the following two issues;
- [ ] The preview of the result is too large. I can't see the actual mention of "indel" in the preview. The preview renders the entire glossary page.
- [ ] The search results aren't correct. "Indel" is used in many more pages (HGVS simple; delins pages on DNA, RNA, and protein; substitution pages on DNA, RNA, and protein; etc). It seems that searching for `indel` doesn't match `(indel)` or `"indel"`.
(note: the term indel will soon be removed in many places, so the term will be less frequent in later code; see #142)

ahwagner:
I looked into this a little; I'm not sure this is something we can address ourselves. It would likely require working with the MkDocs-Material maintainers. [This issue thread ](https://github.com/squidfunk/mkdocs-material/issues/4278) seems to focus on this specific challenge across implementations.

ifokkema:
Thanks for checking this out, Alex!

> I looked into this a little; I'm not sure this is something we can address ourselves. It would likely require working with the MkDocs-Material maintainers. [This issue thread ](https://github.com/squidfunk/mkdocs-material/issues/4278) seems to focus on this specific challenge across implementations.

Right... looks like we'll just have to wait until they rebuild that search feature, then. It'll maybe also solve the current mismatches that I mentioned.


> One workaround, [provided in this comment](https://github.com/squidfunk/mkdocs-material/issues/4278#issuecomment-1562597537), addresses the issue:
> 
> ```css
> .md-search-result .md-typeset {
>   display: -webkit-box;
>   -webkit-line-clamp: 3;
>   -webkit-box-orient: vertical;
>   max-height: 120px; /* <- adjust to fit */
> }
> ```
> 
> However, I think this just makes our results look worse by truncating the actual occurrences of the search term (...)

[...]

PR #193, opened 2024-11-03
Organize fusion recommendations

### LINKED ISSUE #192: Integrate fusion recommendations from SVD-WG007
From the manuscript review:

> _Where is this update documented in the recommendations on the website? A search for “fusion” only shows the community consultation SVD-WG007. If this recommendation is still just in consultation and hasn’t been incorporated into the guidelines yet, that should be noted in the manuscript. If it is part of the current guidelines, I’d expect to find it somewhere in the Recommendations menu._

@reece requested that this work be taken on in the next few days to accommodate reviewer comments.

ahwagner:
Closes #192.

### DIFF
--- a/docs/recommendations/RNA/adjoined_transcript.md
+++ b/docs/recommendations/RNA/adjoined_transcript.md
@@ -9,3 +9,16 @@ Adjoined Transcript: a transcript (RNA molecule) composed of adjoined RNA from t
 ```sh exec="true"
 bin/pull-syntax -c -f docs/syntax.yaml rna.adjoined_transcript
 ```
+
+## Notes
+- Adjoined transcripts are a product of chimeric transcript gene fusions.
+- The adjoined transcript syntax proposed in [Community Proposal SVD WG007](../../consultation/SVD-WG007.md) and adopted
+  here has the following limitations:
+  - This syntax is for two-partner adjoined transcripts only.
+  - This syntax is for RNA sequence only (no use of coding / non-coding DNA reference sequences).
+- This syntax REQUIRES the use of a range (not a single position) for `five_prime_range` / `three_prime_range`.
+- When the adjoined transcript junction but not the entire transcript is analyzed, the outer range bounds SHOULD be
+  specified with `?`, e.g. `NM_152263.2:r.?_775::NM_002609.3:r.1580_?`
+- All adjoined transcripts are described using the same format, irrespective of whether they derive
+  from inter-chromosomal or intra-chromosomal DNA rearrangements (translocation, deletion, inversion)
+  or other mechanisms (trans-splicing).
--- a/docs/recommendations/RNA/adjoined_transcript.md
+++ b/docs/recommendations/RNA/adjoined_transcript.md
@@ -16,7 +16,7 @@ bin/pull-syntax -c -f docs/syntax.yaml rna.adjoined_transcript
   here has the following limitations:
     - This syntax is for two-partner adjoined transcripts only.
     - This syntax is for RNA sequence only (no use of coding (`c.`) / non-coding DNA (`n.`) reference sequences).
-    - Linker sequences are specified using [General Recommendations] for RNA sequence character codes, e.g. `aggcucccuugg`
+    - Linker sequences are specified using [General Recommendations](../general.md) for RNA sequence character codes, e.g. `aggcucccuugg`
 - This syntax REQUIRES the use of a range (not a single position) for `five_prime_range` / `three_prime_range`.
 - When the adjoined transcript junction but not the entire transcript is analyzed, the outer range bounds SHOULD be
   specified with `?`, e.g. `NM_152263.2:r.?_775::NM_002609.3:r.1580_?`
--- a/docs/recommendations/RNA/adjoined_transcript.md
+++ b/docs/recommendations/RNA/adjoined_transcript.md
@@ -7,5 +7,5 @@ Adjoined Transcript: a transcript (RNA molecule) composed of adjoined RNA from t
 ## Syntax
 
 ```sh exec="true"
-bin/pull-syntax -c -f docs/syntax.yaml rna.adjoinedTranscript
+bin/pull-syntax -c -f docs/syntax.yaml rna.adjoined_transcript
 ```
--- a/docs/recommendations/RNA/adjoined_transcript.md
+++ b/docs/recommendations/RNA/adjoined_transcript.md
@@ -27,11 +27,15 @@ bin/pull-syntax -c -f docs/syntax.yaml rna.adjoined_transcript
 ## Examples
 
 - **translocation-derived adjoined transcript**<br>
-    `NM_152263.2:r.-115_775::NM_002609.3:r.1580_*1924` describes an adjoined transcript from a `TPM3::PDGFRB` gene fusion, where nucleotides `r.-115` to `r.775` (reference transcript `NM_152263.2`, _TPM3_ gene) are coupled to nucleotides `r.1580` to `r.*1924` (reference transcript `NM_002609.3`, _PDGFRB_ gene).
+    - **`NM_152263.2:r.-115_775::NM_002609.3:r.1580_*1924`**<br>
+        describes an adjoined transcript from a `TPM3::PDGFRB` gene fusion, where nucleotides `r.-115` to `r.775` (reference transcript
+        `NM_152263.2`, _TPM3_ gene) are coupled to nucleotides `r.1580` to `r.*1924` (reference transcript `NM_002609.3`, _PDGFRB_ gene).
 
 - **deletion-derived adjoined transcripts**
     - **`NM_002354.2:r.-358_555::NM_000251.2:r.212_*279`**<br>
-        describes an adjoined transcript from an `EPCAM::MSH2` gene fusion, where nucleotides `r.-358` to `r.555` (reference transcript `NM_002354.2`, _EPCAM_ gene) are coupled to nucleotides `r.212` to `r.*279` (reference transcript `NM_000251.2`, _MSH2_ gene).
+        describes an adjoined transcript from an `EPCAM::MSH2` gene fusion, where nucleotides `r.-358` to `r.555` (reference transcript
+        `NM_002354.2`, _EPCAM_ gene) are coupled to nucleotides `r.212` to `r.*279` (reference transcript `NM_000251.2`, _MSH2_ gene).
 
     - **`NM_002354.2:r.?_555::guaugauuuuuuaataa::NM_000251.2:r.212_?`**<br>
-        describes an adjoined transcript from an `EPCAM::MSH2` gene fusion, where only the fusion break point has been characterised, showing the insertion of a 17 nucleotide sequence (`guaugauuuuuuaataa`) between two adjoined transcripts.
+        describes an adjoined transcript from an `EPCAM::MSH2` gene fusion, where only the fusion break point has been characterised,
+        showing the insertion of a 17 nucleotide sequence (`guaugauuuuuuaataa`) between two adjoined transcripts.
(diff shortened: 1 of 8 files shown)


ifokkema:
For full integration, I think it would be good to:
- Find some more pages from where we can link to the new adjoined transcripts page. E.g., would it be helpful to add a link from the RNA/deletion page, saying that when a genomic deletion causes two genes to create a single adjoined transcript, that the effect on RNA level should not be described as a deletion, but as an adjoined transcript? Our pages often contain links to other pages that way.
- Use the term "fusion transcript" somewhere, even if it would be to say that these are no longer called fusion transcripts. That way, people who search for it (e.g., this reviewer) will find the page?

docs/syntax.yaml
```diff
@@ -321,8 +320,22 @@ rna:
       reference_nucleotide: the nucleotide at this position in the reference sequence; c
       # ">": the type of change, a substitution
       new_nucleotide: the nucleotide substituting the existing one; g
-
     forms:
       - syntax: sequence_identifier ":" coordinate_type "." position reference_nucleotide ">" new_nucleotide
         examples:
           - NM_004006.3:r.123c>g
+  adjTrx:
```
reece, line 327: @ahwagner `adjacent_transcript` or similar, please, to keep with the more verbose snake_case used everywhere else.

reece:
@ahwagner This is still in draft but marked as closing #192. Intended?

ahwagner:
@reece the intent is that, once completed, this PR will close #192. It is still incomplete and therefore in draft.

ahwagner:
> For full integration, I think it would be good to:
> 
> * Find some more pages from where we can link to the new adjoined transcripts page. E.g., would it be helpful to add a link from the RNA/deletion page, saying that when a genomic deletion causes two genes to create a single adjoined transcript, that the effect on RNA level should not be described as a deletion, but as an adjoined transcript? Our pages often contain links to other pages that way.

This is a good suggestion! I will look for opportunities to cross-link to the adjoined transcript page.

> * Use the term "fusion transcript" somewhere, even if it would be to say that these are no longer called fusion transcripts. That way, people who search for it (e.g., this reviewer) will find the page?

I don't think so. The reviewer searched for "fusion", not "fusion transcript". I think he was confused because the second search result for "fusion" occurs on a page called _Older versions_, so it looks like there is only one result (when there are, in fact, 9 pages on which the term "fusion" appears). Here's what that looks like:

image

And in fact, one of these references is on our general recommendations page, and explicitly links the products of gene fusions to "adjoined transcripts".

> HGVS recommends following the [HGNC guidelines](https://www.genenames.org/about/guidelines/) and the [VICC Gene Fusion Specification](https://fusions.cancervariants.org/en/latest) nomenclature to describe products of gene fusions.
>- HGVS also recommends the use of [adjoined transcripts](https://hgvs-nomenclature.org/stable/recommendations/RNA/splicing/) for precise and unambiguous characterization of chimeric transcripts at the sequence level.

We can also add these clarifications to our response document.

docs/syntax.yaml
ahwagner: addressed in 759b18b

ifokkema:
> > * Use the term "fusion transcript" somewhere, even if it would be to say that these are no longer called fusion transcripts. That way, people who search for it (e.g., this reviewer) will find the page?
> 
> I don't think so. The reviewer searched for "fusion", not "fusion transcript". I think he was confused because the second search result for "fusion" occurs on a page called _Older versions_, so it looks like there is only one result (when there are, in fact, 9 pages on which the term "fusion" appears).

Ah, I see. It's related to #143, then. It would be good to tackle that some time in the near future, as solving that issue will help people use the website better.

docs/recommendations/RNA/adjoined_transcript.md
```diff
@@ -0,0 +1,41 @@
...
+Adjoined Transcript: a transcript (RNA molecule) composed of adjoined RNA from two or more contributing transcripts.
+
+## Syntax
+
+```sh exec="true"
+bin/pull-syntax -c -f docs/syntax.yaml rna.adjoined_transcript
+```
+
+## Notes
+- Adjoined transcripts are a product of chimeric transcript gene fusions.
```
reece, line 14: Can we drop "chimeric transcript"?

docs/recommendations/RNA/delins.md
```diff
@@ -52,16 +52,7 @@ bin/pull-syntax -c -f docs/syntax.yaml rna.delins
...
-
-    - **deletion-derived adjoined transcripts**
-        - **`NM_002354.2:r.-358_555::NM_000251.2:r.212_*279`**<br>
-          describes an adjoined transcript from an `EPCAM::MSH2` gene fusion, where nucleotides `r.-358` to `r.555` (reference transcript `NM_002354.2`, _EPCAM_ gene) are coupled to nucleotides `r.212` to `r.*279` (reference transcript `NM_000251.2`, _MSH2_ gene).
-
-        - **`NM_002354.2:r.?_555::guaugauuuuuuaataa::NM_000251.2:r.212_?`**<br>
-          describes an adjoined transcript from an `EPCAM::MSH2` gene fusion, where only the fusion break point has been characterised, showing the insertion of a 17 nucleotide sequence (`guaugauuuuuuaataa`) between two adjoined transcripts.
+- Example adjoined transcripts from gene fusions (based on [SVD-WG007](../../consultation/SVD-WG007.md)) are discussed in the [adjoined transcript recommendations](adjoined_transcript.md).
```
reece, line 55: I don't think we need to reference WG007 any longer. Instead, just...

- Example adjoined transcripts from gene fusions are discussed in [Adjoined Transcript](adjoined_transcript.md).

reece (changes requested):
Excellent work, @ahwagner . 

I have only one substantive request:

- "adjoined transcript" and "chimeric transcript" seem like the same thing to me. I suspect that there is a shade of meaning between the two that I'm not getting, but I'll also fess up that whatever nuance you're trying to convey is not getting through to me. Since they're so similar, I suspect you could drop one. If not, then please articulate the difference in the definitions.

docs/recommendations/RNA/adjoined_transcript.md
ahwagner: We can replace "chimeric transcript" with "some" to simplify this.

ahwagner: addressed in 2e6dd27

docs/recommendations/RNA/delins.md
ahwagner: I agree; addressed in fed2773

reece:
@ahwagner : The definition of chimeric transcript in the glossary is:

- **chimeric transcript**<br>
  an adjoined transcript derived from two or more genes.

Since an adjoined transcripts are "a product of some gene fusions" (i.e., two or more genes), chimeric transcript seems redundant with adjoined transcript.

This matters in part because we have five uses (at least) of "chimeric transcript"  in the HGVS pages. If chimeric transcript is a deprecated term, we should adjust those uses too. (I would do this, but I'm not confident that I understand distinctions that you might be trying to make.)

ahwagner:
There is a subtle distinction between the two. Chimeric transcripts are not a deprecated term. I simplified the language in 2e6dd27 because I think the distinction was more distracting than helpful for that statement.

Gene fusions are broadly split into two classes: _regulatory fusions_ and _chimeric transcript fusions_. Chimeric transcript fusions MUST involve 2 or more genes. A related, common class of (non-gene-fusion) variant is the internal tandem duplication, which results in RNA molecules that may be concisely represented as adjoined transcripts from the _same_ gene. Our nomenclature structure that encompasses both types of events, as well as other types of spliced multi-transcript events that don't meet the stricter definition of a chimeric transcript, is the Adjoined Transcript.

I looked through other usages of "chimeric" in code and most places use it with "chimeric transcript fusions", which is a proper concept name. One place (the SVD-WG007 consultation admonition) explicitly indicates this is a separable concept from "adjoined transcripts" and provides some hints about it. I can expand on this in the recommendations somewhere if helpful.

reece (approved):
lgtm. Thanks for replying to my comment (in Conversation) re: the difference between adjoined and chimeric transcripts.

reece:
@ifokkema @marinadistefano @jfjlaros : Given the size of the adjoined transcripts PR, I think you should have an opportunity to weigh in. I'll plan to merge this PR Tue morning (US Pacific Time) unless I hear concerns, or sooner if it gets a second approval.

docs/recommendations/general.md
```diff
@@ -64,7 +64,7 @@ Make sure you have also seen the [Basics](../background/basics.md), explaining t
 - HGVS recommends following the [HGNC guidelines](https://www.genenames.org/about/guidelines/) and the [VICC Gene Fusion Specification](https://fusions.cancervariants.org/en/latest) nomenclature to describe products of gene fusions.
     - The HGNC recommendations include using a `GENESYMBOL1::GENESYMBOL2` syntax for gene-level fusion descriptions, and a `GENESYMBOL1-GENESYMBOL2` syntax for read-through transcripts.
     - The VICC nomenclature extends the HGNC recommendations to include a terminology, information model, and nomenclature for gene-level and exon-level representation, with components for disambiguating regulatory fusions from chimeric transcript fusions.
-    - HGVS also recommends the use of [adjoined transcripts](RNA/splicing.md) for precise and unambiguous characterization of chimeric transcripts at the sequence level.
+    - HGVS also recommends the use of [adjoined transcripts](RNA/adjoined_transcript.md) for precise and unambiguous characterization of chimeric transcripts at the sequence level.
```
ifokkema, line 67: This change prompted me to check the splicing page because I was wondering why we linked to that page to explain adjoined transcripts. I see that the splicing page still lists an example of an adjoined transcript and links to SVD-WG007. Maybe we should remove that? I don't see why the splicing page should have that example, nor why somebody looking for adjoined transcripts would check the splicing page :thinking:

docs/recommendations/protein/delins.md
```diff
@@ -28,9 +28,9 @@ bin/pull-syntax -c -f docs/syntax.yaml aa.delins
     - **encode a translation stop codon** in the inserted sequence are described as deletion-insertion of this sequence, not as a deletion-insertion removing the entire C-terminal amino acid sequence.
     - encode an open reading frame which **after** the inserted sequence shifts to another reading frame, are described as a [frameshift](frameshift.md).
     - **start N-terminal (5') of and including** the translation termination (stop) codon, are described as [frameshift](frameshift.md).
-- fusion proteins represent a special case of deletion-insertion variant.
-  The fusion break point is described using **`::`**.<br>
-  **NOTE**: to avoid confusion, HGVS recommends to follow the [HGNC guidelines](https://www.genenames.org/about/guidelines/) to describe products of gene translocations or fusions (format `GENESYMBOL1::GENESYMBOL2`) and readthrough transcripts (format `GENESYMBOL1-GENESYMBOL2`).
```
ifokkema, line 33: This same sentence is still on the RNA delins page. For consistency, shouldn't both sentences be adjusted? However, I'm wondering about the examples given (`GENESYMBOL1::GENESYMBOL2` and `GENESYMBOL1-GENESYMBOL2`). They are now removed; are they not relevant here? The adjoined transcripts page only lists the `GENE::GENE` example, not the `GENE-GENE` readthrough format. The RNA splicing page and the general page still list both examples. Perhaps we should clarify somewhere how adjoined transcripts are different from readthrough transcripts?

ifokkema (commented):
I realize this has been merged already, but the deadline was too short for me to be able to review it on time. I have a few concerns; some changes create inconsistency and raise doubts.

ifokkema:
While reviewing, I also thought of a question. What if two genes are adjoined in such a way that the adjoined transcript contains part of an intronic sequence? If we allow a linker sequence written out, can we also use locations to indicate the linker sequence if it's long? That would make it similar to insertions of sequence found elsewhere. If so, what would that look like? A fairly complex description, I'd reckon. I would recommend adding the answer to this question as a Q&A on the adjoined transcript page.

ifokkema:
I also realize that some PRs are getting merged with the commits squashed. I'm not a fan of this, as it rewrites the history and essentially removes commits made by the author of the PR. Now and then, I go back in history to see how a page was constructed, and this won't work as well if PRs are squashed. This may be a personal preference or just something done out of habit; perhaps we can discuss and come to a consensus?

PR #198, opened 2024-11-06
Add the LOVD HGVS syntax checker to the software list.

ifokkema:
Add the LOVD HGVS syntax checker to the software list. Also added `library/php` as a valid type.


reece (approved):
lgtm

PR #207, opened 2024-12-20
updated pre-commit hooks to 5.0.0

reece:


reece:
Fair point.

In this case, it was mostly to silence deprecation warnings about a change
in the precommit API.

I'll comment next time

PR #208, opened 2024-12-20
Updated 2024 citation with published URLs

reece:


ifokkema (approved):
The change looks good to me!

I think it would be nice, in the near future, to consider a plugin like [this one](https://github.com/JakubAndrysek/mkdocs-open-in-new-tab) to make sure external links open in a new tab. That's not for this PR, but I ran into this issue again while testing.

PR #210, opened 2024-12-27
open external links in new tab and indicate with icon next to link

reece:
@ifokkema asked for it!

This PR causes external links to be opened in a new tab and denotes such links with a conventional icon adjacent to the link, as shown below.
image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -45,7 +45,7 @@ extra:
       link: https://github.com/HGVSNomenclature/hgvs-nomenclature
 
 extra_css:
-  - stylesheets/extra.css?v=1
+  - stylesheets/extra.css
 
 # enabling strict above (or mkdocs build --strict) causes warnings to become fatal
 validation:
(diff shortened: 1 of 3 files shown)


mkdocs.yml
```diff
@@ -43,7 +45,7 @@ extra:
       link: https://github.com/HGVSNomenclature/hgvs-nomenclature
 
 extra_css:
-  - stylesheets/extra.css
+  - stylesheets/extra.css?v=1
```
ifokkema, line 48: This change doesn't seem to do anything, I think... :thinking:

ifokkema (approved):
Sweet, thanks!
Besides one change that may have been an unintended commit (adding "?v=1" to the CSS URL, possibly to force the CSS to be reloaded while testing), it looks good to me!

mkdocs.yml
reece: You're right of course. Will fix before merging.

PR #211, opened 2025-01-03
Enable fontawesome

reece:
#210 didn't work as expected because I had font-awesome fonts installed locally, so didn't see that the fonts were not being loaded.

This PR adds fontawesome css to pages so that fontawesome code points are available (e.g., `\f08e`) but not the glyph names (`:smile:`)


reece:
@ahwagner : I reproduced the issue, then squashed it with this PR. 
Please confirm at https://hgvs-nomenclature--211.org.readthedocs.build/211/ .

ahwagner (approved):
LGTM

PR #213, opened 2025-01-27
Update standards.md

jtdendunnen:
IUPAC does not use lower case nor a "u" for describing nucleosides on RNA level. HGVS nomenclature follows IUPAC so this needs to be corrected.

### DIFF
--- a/docs/background/standards.md
+++ b/docs/background/standards.md
@@ -44,21 +44,21 @@ On DNA and RNA level, HGVS nomenclature follows the _Nomenclature for Incomplete
 
 | Symbol |   Meaning    |           Description           |
 |:------:|:------------:|:-------------------------------:|
-|   a    |      a       |            Adenosine            |
-|   c    |      c       |            Cytidine             |
-|   g    |      g       |            Guanosine            |
-|   u    |      u       |             Uridine             |
-|   b    |  c, g or u   | not-a (b follows a in alphabet) |
-|   d    |  a, g or u   | not-c (d follows c in alphabet) |
-|   h    |  a, c or u   | not-g (h follows g in alphabet) |
-|   k    |    g or u    |              keto               |
-|   m    |    a or c    |              amino              |
-|   n    | a, c, g or u |               any               |
-|   r    |    a or g    |             purine              |
-|   s    |    g or c    | strong interaction (3 H-bonds)  |
-|   v    |  a, c or g   |       not-u (v follows u)       |
-|   w    |    a or u    |  weak interaction (2 H-bonds)   |
-|   y    |    c or u    |           pyrimidine            |
+|   A    |      A       |            Adenosine            |
+|   C    |      C       |            Cytidine             |
+|   G    |      G       |            Guanosine            |
+|   T    |      T       |             Uridine             |
+|   B    |  C, G or T   | not-A (B follows A in alphabet) |
+|   D    |  A, G or T   | not-C (D follows C in alphabet) |
+|   H    |  A, C or T   | not-G (H follows G in alphabet) |
+|   K    |    G or T    |              Keto               |
+|   M    |    A or C    |              aMino              |
+|   N    | A, C, G or T |               aNy               |
+|   R    |    A or G    |             puRine              |
+|   S    |    G or C    | strong interaction (3 H-bonds)  |
+|   V    |  A, C or G   |       not-U (V follows U)       |
+|   W    |    A or T    |  Weak interaction (2 H-bonds)   |
+|   Y    |    C or T    |           pYrimidine            |
 
 <a id="geneticcode"></a>
 


docs/background/standards.md
```diff
@@ -44,21 +44,21 @@ On DNA and RNA level, HGVS nomenclature follows the _Nomenclature for Incomplete
...
-|   n    | a, c, g or u |               any               |
-|   r    |    a or g    |             purine              |
-|   s    |    g or c    | strong interaction (3 H-bonds)  |
-|   v    |  a, c or g   |       not-u (v follows u)       |
-|   w    |    a or u    |  weak interaction (2 H-bonds)   |
-|   y    |    c or u    |           pyrimidine            |
+|   A    |      A       |            Adenosine            |
+|   C    |      C       |            Cytidine             |
+|   G    |      G       |            Guanosine            |
+|   T    |      T       |             Uridine             |
```
jfjlaros, line 50: U is a valid IUPAC code.

jtdendunnen: Yes, but it is not used in describing nucleotides. So what do you suggest?, we add a note stating this?

jtdendunnen: I had another look at the IUPAC pages and suggest we add a link to https://iupac.qmul.ac.uk/misc/naabb.html#symbol for the DNA/RNA table and https://iupac.qmul.ac.uk/AminoAcid/AA1n2.html for the amino acid table. In addition, we should mention "as recommended by IUPAC, symbols will not discriminate between DNA and RNA, and the symbol T will be employed at all positions where U might appear in the RNA" (link to https://iubmb.qmul.ac.uk/misc/naseq.html#500). I wanted to update the page and add these, but I was (unexperienced user I assume) not able to find the page with the changes I suggested.

jfjlaros: > "as recommended by IUPAC, symbols will not discriminate between DNA and RNA [...]"

The page referred to is not from IUPAC but from IUBMB. This recommendation is specifically about incompletely specified (ambigious) bases in sequences. In this context, it makes sense to make no distinction between `U` and `T` to keep the number of ambiguous codes manageable (e.g., both (`T` or `C`) and (`U` or `C`) are encoded as `Y`).

jfjlaros: The recommendation you are referring to:

> Thus it is important to note from the outset that the recommended symbols will not discriminate between DNA and RNA, and the symbol T will be employed at all positions where U might appear in the RNA.

is from https://iubmb.qmul.ac.uk/misc/naseq.html#100, which is not a IUPAC page.

My point is that the wrong arguments are used to deprecate the use of `U`. My suggestion would be to either drop this merge request, or to go through the formal change procedure.

jfjlaros:
As I pointed out previously, there is no conflict with existing standards, at least not with the one relevant for this discussion (IUPAC). As I see it, there was no error made by the committee and therefore there is nothing to correct.

Furthermore, the use of `u` (lowercase) has been part of the recommendations since at least 2002. We cannot silently change this without going through the proper procedure.

jtdendunnen:
As far as I understand IUPAC was a subcommittee of the Nomenclature Committee of the International Union of Biochemistry (NC-IUB). The NC-IUB published the latest recommendations in 1986:

Nomenclature for incompletely specified bases in nucleic acid sequences. Recommendations 1984. Nomenclature Committee of the International Union of Biochemistry (NC-IUB). Proc Natl Acad Sci USA. 1986, Jan;83(1):4-8. doi: 10.1073/pnas.83.1.4.

The recommendation includes, as mentioned on https://iubmb.qmul.ac.uk/misc/naseq.html#500:
Thus it is important to note from the outset that the recommended symbols will not discriminate between DNA and RNA, and the symbol T will be employed at all positions where U might appear in the RNA.

The final table is at (as linked on our HGVS nomenclature web site hgvs-nomenclature.org/stable/background/standards/:

5. DISCUSSION
The present nomenclature, summarised in Table 1, has been formulated to deal with incomplete specification of bases in nucleic acid sequences.

In this table the use of "U" is not mentioned. 

My proposal therefore is to change the the page on hgvs-nomenclature.org/stable/background/standards/ by:

- removing the RNA table
- changing the DNA heading to "DNA and RNA"
- adding (top) the sentence:
The recommended symbols will not discriminate between DNA and RNA, and the symbol T will be employed at all positions where U might appear in RNA.

ifokkema:
Note that the page that Johan links to, which mentions there should not be a discrimination between DNA and RNA, is a page named "Nomenclature for Incompletely Specified Bases in Nucleic Acid Sequences". This does not apply to normal usage.

ifokkema:
I cannot find any recommendation on the [official page](https://iupac.qmul.ac.uk/misc/naabb.html) against using U in RNA sequences or in favor of using T in those cases, but I do agree that there is no reason specified for using lowercase letters.

PR #228, opened 2025-03-25
Highlight negative examples as such.

ifokkema:
The HGVS nomenclature website uses negative examples to show what not to do. These examples are almost always labeled as invalid, so they are rendered in red, indicating visually to the reader that these variant descriptions are invalid.

A [recent analysis](https://github.com/HGVSnomenclature/hgvs-nomenclature/discussions/212) showed 89 invalid DNA descriptions on the HGVS website that were not labeled as invalid. This fixes those descriptions that are (IMO) clearly invalid examples, by labeling them as invalid examples. The descriptions themselves are not edited.


jfjlaros (approved):
Although I do not agree with the term "invalid", the labeling seems to be consistently applied to non-canonical descriptions.

reece (approved):
Looks great to me. Thanks for attending to these, @ifokkema !

ifokkema:
Thanks for reviewing @jfjlaros and @reece, and merging!
But would it be possible to use a simple merge? Or perhaps let the author of the PR decide how the PR should be merged? The squash-and-merge disturbs my workflow quite a bit... Besides my commit history being lost, it also requires me to force-delete my local branch which requires me to double-check things. Maybe we can come up with a standard workflow and add it to the dev docs?

PR #229, opened 2025-03-31
Fix several typos on the website and in examples

ifokkema:
### Fix several typos on the website and in examples
- Only complex insertions need square brackets.
- Fix old syntax; `del(15)` has been replaced by `delN[15]`.
- Fix typos in a complex variant, although it's still invalid.
  - I noticed the variant because the positions were in the wrong order. However, based on the text, I could determine what was most likely meant.
  - However, I had to fix the text as well, as that contained a typo, too.
  - Finally, the variant still isn't valid since that reference sequence doesn't have exon endings at those positions. I have no clue what the source of this description is.
- Fix two forgotten prefixes (`c.` and `g.`).
- Fix some typos on the grammar page.


jfjlaros (approved):
I am not sure the first and the fifth change are really needed, but I have no objection either.

PR #230, opened 2025-05-15
Fix `X` to `Xaa`

ifokkema:
### Fix `X` to `Xaa` in protein descriptions.
- We recommend `Xaa` over X, but we didn't use it anywhere ourselves. This replaces all uses of `X` with `Xaa`.
- Add another reminder to use `Xaa` instead of `X`.


jfjlaros (approved):
This looks good to me.

PR #231, opened 2025-05-20
Clarify that repeats have to be divisible by three only in the CDS.

ifokkema:
Clarify that repeats have to be divisible by three only in the CDS. UTRs and introns of coding sequences are excluded from this requirement. Also added an example.

This is task 13 from the HVNC meetings. Note that during the meeting, we only discussed introns. I believe that the UTRs are meant to be excluded as well. This change will have to be reviewed in the upcoming HVNC meeting in June. @ahwagner, could you please add the review of this PR to the agenda when you create it? Thank you!


jfjlaros (approved):
This is in line with an upcoming proposal to drop the restriction on repeats in the CDS.

PR #233, opened 2025-07-28
Remove all links to SVD-WG010, since it's been rejected.

ifokkema:


jfjlaros (approved):
This looks fine to me.

As a side note, we seem to have a lot of repetition in our pages. Perhaps we should think about using links.

ifokkema:
> > [...] but they are visible to the users without them having to click to a different page, read, and go back.
> 
> Maybe we could do something with mouseover events.

Are the notes in the way for you? They're not for me, so I prefer having them there clearly to prevent people from not seeing them or overlooking them. Although the text is sometimes a bit confusing, I also find that people sometimes just don't read properly. Hiding the text from them will just make things worse, I think. We can, as an alternative, list short reminders with links or mouseovers to the full text/explanation about the issue?

gostachowiak:
@ifokkema 
The recommendations/general page still has this: 

`NOTE: the SVD-WG is preparing a proposal to modify this recommendation. To apply the current rule one needs to know whether the two variants are in a coding sequence and affecting one amino acid. Recommendations should be general. The new recommendation will be: two variants separated by less than two nucleotides should be described as a "delins".`

Should this also be deleted?

PR #252, opened 2026-05-04
Update banner and call for members

ahwagner:


docs/call-for-members.md
```diff
@@ -1,40 +1,34 @@
...
-of all sequence variants on the DNA, RNA, and protein level (see
-[https://hgvs-nomenclature.org](https://hgvs-nomenclature.org)). The committee's
-[Terms of
-Reference](https://mcusercontent.com/8b89ab9930fa0d9d5f17116d3/files/95e48a7f-dce5-bbf8-6cef-226a37436464/HVNC_ToR.01.doc)
-can be found on the HUGO website
+Standards Committees, is looking for new members for the 2026-2030 term. The
+main activity of the committee is the "HGVS nomenclature", the standard for the
+unambiguous description of all sequence variants at the DNA, RNA and protein level
+(see [www.HGVS-nomenclature.org](https://hgvs-nomenclature.org)). The committee's
+Terms of Reference can be found on the HUGO website
```
ifokkema: The terms of reference are now located here, if you want to link to that directly:
https://www.hugo-international.org/wp-content/uploads/2025/12/HVNC_TermsofReference-202504.pdf

The current link shows the standards page, and we're at the bottom of it. I realize there is no better way to link to our section, as no link target was made for this, but currently, I got a bit confused and thought maybe the link was wrong. Others may have the same idea. I would personally recommend linking to the PDF; we should then ask somebody at HUGO to create a link target so that we can directly link to the HVNC section on the HUGO website in the future.

docs/call-for-members.md
```diff
@@ -1,40 +1,34 @@
 # Call for Members
 
-!!! warning "Expired"
-
-    This Call for Members has expired.
-    If you are interested in joining the HVNC in the future,
-     become a member of our [HGVS Nomenclature group](https://groups.google.com/g/hgvs-nomenclature)
```
ifokkema, line 7: Seeing this removed, have we announced the call for members there? I don't think so, so we probably should.
It could also be wise to save this text as a comment, so it's easily retrievable when we close the call after June 1st.

ifokkema (approved):
Just minor comments, as suggestions. So, approving nonetheless.

PR #260, opened 2026-07-03
Fix example, numbering, and markup errors

gostachowiak:
I used AI to identify these issues, but I reviewed all of them and they seem correct and obvious to me.

### DIFF
--- a/docs/background/simple.md
+++ b/docs/background/simple.md
@@ -12,10 +12,10 @@ More details, on all subjects, are available elsewhere on the HGVS nomenclature 
 
 The format of a complete variant description is **"reference : description"** (spaces added for clarity only), e.g.;
 
-- `NM_004006.2:c.4375C>T`
+- `NM_004006.3:c.4375C>T`
 - `NC_000023.11:g.32389644G>A`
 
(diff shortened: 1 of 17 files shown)


docs/background/simple.md
```diff
@@ -12,10 +12,10 @@ More details, on all subjects, are available elsewhere on the HGVS nomenclature
 
 The format of a complete variant description is **"reference : description"** (spaces added for clarity only), e.g.;
 
-- `NM_004006.2:c.4375C>T`
+- `NM_004006.3:c.4375C>T`
```
gostachowiak, line 15: NM_004006 appears 10 times on this page.  Some are .2 and some are .3.  I confirmed that .3 is consistent with all claims about the transcript on the page (namely that NM_004006.3:c.4375C>T = p.(Arg1459Ter))

docs/recommendations/DNA/complex.md
```diff
@@ -124,7 +124,7 @@ HGVS, therefore, recommends to describe translocations exclusively using a "deli
       for ISCN `del(X)(q21.31q22.2)`<br>
       _(within a chromosome, breakpoint not sequenced)_
 
-    - **`NC_000022.11:g.pter_(12200001_14700000)del::(37600001_410000000)_qterdel`**<br>
+    - **`NC_000022.11:g.pter_(12200001_14700000)del::(37600001_41000000)_qterdel`**<br>
```
gostachowiak, line 127: chr22 does not have position 410000000.  It does have position 41000000 which is close to the q13.1 band mentioned below

PR #266, opened 2026-08-11
docs: align the 3'rule exon-junction exception with the RNA pages

nh13:
`recommendations/general.md` lists three reference-sequence types as taking the exon/exon junction exception to the 3'rule:

> **exception**: deletions/duplications around exon/exon junctions using **c.**, **r.** or **n.** reference sequences

Two RNA pages state the opposite for one of them. `RNA/deletion.md`:

> the exception to the 3'rule for deletions around exon/exon junctions … **does not apply** when describing variants based on an RNA reference sequence

and `RNA/duplication.md` says the same for duplications.

A tool must either shift or not shift at an exon junction on an `r.` description, and the two pages give opposite instructions.

This drops `r.` from the `general.md` list and adds a sub-note pointing at the two RNA pages, so the three statements agree. It follows the RNA pages because they state the restriction twice and unambiguously; if the intended reading is the other way round, the fix belongs on those pages instead.


jfjlaros:
I think the following statement in the [general recommendations](https://hgvs-nomenclature.org/stable/recommendations/general/#general-recommendations) should be leading here.

> descriptions on RNA/protein level should describe the changes observed on that level (RNA/protein) and not try to incorporate any knowledge regarding the change on DNA-level

To me, this implies that the 3' rule always applies, it only applies on different (perhaps constructed) reference sequences. For example, for RNA (`r.`) descriptions, we should not use any knowledge about exon structure, therefore we should shift over exon-exon boundaries when needed; we use a constructed RNA reference sequence (remove up- and downstream regions and introns and transcribe to RNA) before we apply the 3' rule. Likewise, for protein (`p.`) descriptions, we should not use any knowledge about codon content, so we should apply the 3' rule when a protein repeat is involved, even if this is not a repeat on RNA level; we use a constructed protein sequence (remove 3' and 5' UTRs and translate to protein) before we apply the 3' rule.

One source of confusion is the fact that we allow both the `r.` as well as the `c.` coordinate system when describing a variant on an RNA reference sequence (e.g., NM_002001.3). Since this reference sequence does not contain introns, the effect of the 3' rule is the same for both coding DNA descriptions (`c.`) and RNA descriptions (`r.`), while they may differ when using a genomic reference sequence, in which case the DNA description may be shifted into an intron (or not shifted at all), while the RNA description may be shifted over an exon-exon boundary.

Long story short, I think the 3' rule should be described on a more fundamental (sequence oriented) level to fully resolve this issue.

nh13:
@jfjlaros Thank you for the explanation, this really made it click for me:
> One source of confusion is the fact that we allow both the r. as well as the c. coordinate system when describing a variant on an RNA reference sequence...

Since the (mRNA) reference sequence does not contain introns, the effect of the 3' rule is the same for both coding DNA descriptions (c.) and RNA descriptions (r.), while they may differ when using a genomic reference sequence.

So on an mRNA reference there is no junction for a shift to stop at, and c./r. agree; on a genomic reference the run simply ends at the intron, so the same variant shifts differently.   The exception (general.md:44 states it for c., r. or n. reference sequences) makes a transcript description stop at a boundary that is not present in its own reference sequence.

Also, if you project the mRNA-normalized form back to g., you don't land where normalizing on the genomic reference would have put you. So normalization and projection do not commute. The exception looks like it gives up level-independence at the junction in order to keep the projection consistent and the description canonical.

If that all checks out, then I agree the 3' rule should be stated at a more fundamental, sequence-oriented level.  Should we also name which sequence is authoritative when the levels disagree, since a fully level-independent rule and a projection-consistent one cannot both hold at a junction?

ifokkema:
@jfjlaros 
> Likewise, for protein (`p.`) descriptions (...); we use a constructed protein sequence (remove 3' and 5' UTRs and translate to protein) before we apply the 3' rule.

A minor detail, but this isn't true when protein reference sequences are used. Then, the reference sequence should be leading (e.g., an NP).


> One source of confusion is the fact that we allow both the `r.` as well as the `c.` coordinate system when describing a variant on an RNA reference sequence (e.g., NM_002001.3). Since this reference sequence does not contain introns, the effect of the 3' rule is the same for both coding DNA descriptions (`c.`) and RNA descriptions (`r.`), while they may differ when using a genomic reference sequence, in which case the DNA description may be shifted into an intron (or not shifted at all), while the RNA description may be shifted over an exon-exon boundary.

I disagree here; I don't think `NM_123456.1:c.100del` and `NC_000001.10(NM_123456.1):c.100del` should be 3' shifted differently. The way I interpret the 3' shifting rules, they are based on the coordinate system used. That is done to prevent this exact issue:

> Also, if you project the mRNA-normalized form back to g., you don't land where normalizing on the genomic reference would have put you. So normalization and projection do not commute.

Apart from genes located on the reverse strand, which cause the genomic and cDNA representations of the same variant to shift further apart, it was deemed desirable that the g. notation and c. notation align and keep referring to the same location:

> the 3' rule is not applied when there is a deletion/duplication around exon/exon junctions with identical nucleotides flanking the junction, where shifting the variant 3' would place it in the next exon. Projecting the variant from c. back to g. positions (genomic) would then lead to an incorrect genomic position (in the wrong exon).

[(source)](https://hgvs-nomenclature.org/stable/background/numbering/#DNAc)

As such, `NM_123456.1:c.100del` and `NC_000001.10(NM_123456.1):c.100del` should 3' shift using the same rules, hence, not crossing over to the next exon. However, the r. coordinate system does not handle introns; r.100+1 does not exist, regardless of what reference sequence is used. As such, when using the r. coordinate system, one _should_ cross over exon/exon borders, which are purely virtual for r. descriptions.


@nh13 
> If that all checks out, then I agree the 3' rule should be stated at a more fundamental, sequence-oriented level. Should we also name which sequence is authoritative when the levels disagree, since a fully level-independent rule and a projection-consistent one cannot both hold at a junction?

When using transcript reference sequences with a genomic context (sometimes referred to as "genomic transcripts", e.g., the NC(NM) construct), the level-independent rule _actually is_ projection-consistent. This is already required for describing intronic variants at the c. level, so it is not very uncommon (although there are issues that we are currently trying to figure out). Anyway, I think this idea will create far more problems than it would solve, and I would be strongly against it (but I obviously do not speak for the entire committee). The coordinate systems and reference sequences are separate entities, and one should not confuse them as merged.

jfjlaros:
@nh13 

> So normalization and projection do not commute.

Indeed, if by projection you mean transcription or translation. This level independence has the desirable property that different causes with the same effect have the same effect description.

> Should we also name which sequence is authoritative when the levels disagree, since a fully level-independent rule and a projection-consistent one cannot both hold at a junction?

The only authoritative description should be that of the actual observation, all (back)transcriptions or (back)translations are merely predictions. I do not see how descriptions on different levels can disagree.

@ifokkema 

> [...] `NM_123456.1:c.100del` [...]

As stated before in other discussions, in my opinion, a coding DNA coordinate system in combination with an RNA reference sequence does not make sense and should be removed from the recommendations together with the exception you quoted. This exception is also in conflict with the more general recommendation I referred to earlier.

If RNA sequences were mapped to an RNA reference sequence and variant calling was performed on this same reference sequence, then obviously the 3' rule should apply as in other cases.

ifokkema:
@jfjlaros 
> > [...] `NM_123456.1:c.100del` [...]
> 
> As stated before in other discussions, in my opinion, a coding DNA coordinate system in combination with an RNA reference sequence does not make sense and should be removed from the recommendations (...)

Are you suggesting that the, by far, most common HGVS expression in reports, databases, and the literature, be removed from the HGVS Nomenclature? What reference sequence are you suggesting to use for c. variants? Note that, even though NM reference sequences are labeled as "mRNA" reference sequences, they contain cDNA. Hence the use of "T" instead of "U". It's mRNA -> RT-PCR -> cDNA -> sequencing -> NM reference sequence. Nowadays, since MANE, they are even fully based on the genomic DNA sequence and, therefore, fully gDNA. Should it then not be usable for DNA-based variant descriptions?

Note also that, albeit less so now, c. coordinates were by far the most commonly observed variant. PCR fragments (DNA) are sequenced and aligned to the NM reference sequence; therefore, the variant calling is done at the cDNA level, and `NM_123456.1:c.100del` is then the first observed variant. Any genomic variant description would then be a prediction, as no alignment with the genomic reference sequence was done.

If you then argue that, indirectly, the genomic DNA is sequenced, sure. Indirectly. If you really look into detail, everything is done indirectly. There is no true RNA reference sequence. There aren't even true protein reference sequences; they are just virtual reference sequences based on the translated NMs. We're constantly translating one thing into the other thing. NC:g. may be the only variant description that fully makes sense, strictly speaking, but are only "purely" generated by very few NGS platforms. Normally, genomic DNA is also amplified and, therefore, sequenced indirectly in, relatively speaking, short reads. If I understand correctly what you're trying to correct, there is no correcting it, really.


> (...) together with the exception you quoted. This exception is also in conflict with the more general recommendation I referred to earlier.

As we're always having so many discussions in parallel, I don't know what you mean. What are you referring to here?


> If RNA sequences were mapped to an RNA reference sequence and variant calling was performed on this same reference sequence, then obviously the 3' rule should apply as in other cases.

RNA sequencing is relatively new and still relatively uncommon; normally, RT-PCR is performed, cDNA is sequenced, and the sequence is aligned to the DNA-based NM. In reality, any NM:r variant description is just a translation. What you're suggesting, if I understand correctly, is to add a rule that only "pure" variant descriptions are used (reference sequences and coordinate system fully matching the sequenced molecule). That would require an extensive redesign of the entire HGVS Nomenclature, would invalidate the vast majority of reported variant descriptions, would not be possible for by far most data out there, and, therefore, would never make it through the community consultation step.

jfjlaros:
> Are you suggesting that the, by far, most common HGVS expression in reports, databases, and the literature, be removed from the HGVS Nomenclature?

Indeed.

> What reference sequence are you suggesting to use for c. variants?

Genomic sequences (e.g., NC, NG, NT, etc.). How reference sequences were/are created is purely a technical detail we should not concern ourselves with. The techniques to create these reference sequences can (and probably will) change in the future, maybe they even already have changed for the more recent transcripts.

> What are you referring to here?

I am referring to the exception to the 3' rule you quoted in this conversation.

> [...] reference sequences and coordinate system fully matching the sequenced molecule [...]

I am not suggesting that. I merely pointed out that in the situation I presented, the 3' rule should be applied just as we would in other situations.

nh13:
@ifokkema — that answers my question rather than needing a new rule. If `c.` keys on the coordinate system and never crosses a junction, then projection stays consistent and there is nothing to arbitrate; my premise only bites if `c.` is level-independent, and you're saying it isn't.

I can also name where my confusion came from. `general.md` couples the two: the accepted prefixes are given as "`c` for a coding DNA reference sequence", "`g` for a linear genomic reference sequence", and so on — each prefix assigned to a *type of reference sequence*. Read straight, the prefix tells you the reference type, so it wasn't obvious to me that shifting could key on the prefix while the sequence underneath varies. `NM_…:c.` and `NC_…(NM_…):c.` share a prefix and differ in whether introns are present, which is the case the coupling doesn't decide.

On the PR itself: you both agree `r.` should cross exon/exon borders, since they are virtual at that level. That is all this change does — it drops `r.` from the `general.md` list and points at the two RNA pages, so the three statements agree. It takes no position on `c.`, or on where the 3' rule is best stated.

If it would help, I'm happy to open separate issues to track the two broader threads: where the 3' rule is best stated, and the prefix/reference-sequence relationship.

docs/recommendations/general.md
```diff
@@ -40,7 +40,8 @@ Make sure you have also seen the [Basics](../background/basics.md), explaining t
 - **3'rule**: for all descriptions, the most 3' position possible of the reference sequence is arbitrarily assigned to have been changed.
     - the 3'rule also applies for changes in single residue stretches and tandem repeats (nucleotide or amino acid).
     - the 3'rule applies to ALL descriptions (genome, gene, transcript, and protein) of a given variant.
-    - **exception**: deletions/duplications around exon/exon junctions using **c.**, **r.** or **n.** reference sequences (see [Numbering](../background/numbering.md#DNAc)).
+    - **exception**: deletions/duplications around exon/exon junctions using **c.** or **n.** reference sequences (see [Numbering](../background/numbering.md#DNAc)).
+        - this exception does **not** apply when describing variants based on an RNA reference sequence (see [RNA Deletion](RNA/deletion.md) and [RNA Duplication](RNA/duplication.md)).
```
ahwagner:
```suggestion
```

ahwagner (changes requested):
Please remove this line, then okay to merge per committee.

@ifokkema volunteered to drive.

nh13:
Thanks @ahwagner, removed that line and pushed. @ifokkema, it's yours to merge when ready.

docs/recommendations/general.md
nh13: Good suggestion, I've removed it!

ifokkema (approved):

ifokkema:
@nh13
Thanks, Nils! The issue about the coordinate system and reference sequence being linked by our poor wording is something that also needs to be fixed, but it will need to be discussed separately.

PR #272, opened 2026-09-17
Hide call-for-members page from nav

ahwagner:
## Summary
`docs/call-for-members.md` was being swept into the site nav by the bare `...` rest wildcard in [docs/.pages](docs/.pages), even outside of an active call-for-members cycle.

This moves it into its own directory, `docs/call-for-members/index.md`, with a sibling `.pages` file setting `hide: true`. That's the `mkdocs-awesome-pages-plugin`'s documented mechanism for hiding a whole directory from the nav while still building and serving its page(s) — so the page stays reachable at the same URL (`/call-for-members/`), which matters since it's still linked directly from the site's announcement banner and from commented-out "Join Us" notes in `hvnc.md`/`index.md` (updated here to point at the new path).

An earlier attempt used a negative regex/glob directly on the `...` wildcard in `docs/.pages` to exclude just this one filename. It worked, but as a side effect broke the wildcard's matching generally, which silently dropped other unrelated pages (e.g. `software/`, `versions/*`) from the nav too. This directory+`hide: true` approach only affects the one page.

## Testing
- `mkdocs build --strict` succeeds with no new warnings/errors.
- Verified `docs/call-for-members/index.html` is still built and listed in the sitemap.
- Verified the page no longer appears in the rendered nav on the homepage or hvnc page, while `software/`, `versions/*`, etc. remain in the nav as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ifokkema (approved):
LGTM!

PR #273, opened 2026-09-17
Stop tracking generated docs/versions/*.md stub files

ahwagner:
## Summary
`docs/versions/21.0.md` and `21.1.md` are committed as blank stub placeholders, but [simple_hooks/generate_release_notes.py](simple_hooks/generate_release_notes.py) (wired up via the `mkdocs-simple-hooks` plugin's `on_pre_build` hook in [mkdocs.yml](mkdocs.yml)) overwrites them with freshly-fetched GitHub release note content on every local `mkdocs build`/`mkdocs serve`. Since they were tracked in git, this made them show up as locally modified after essentially any build — pure noise with no actual content to commit.

[docs/versions/.gitignore](docs/versions/.gitignore) already contains `*.md`, which is clearly meant to cover exactly this generated content, but it has no effect on files that were already tracked before the ignore rule was added.

This PR untracks the two files (`git rm --cached`) so the existing ignore rule finally applies to them, same as it already would for any brand-new major.minor series file.

## Why this is safe
The stub files were originally added (see `f86c5bc`) to prevent nav errors, presumably out of concern that `docs/versions/.pages` (which explicitly lists `21.1.md`/`21.0.md` in its nav) would fail to resolve them if they didn't exist yet.

That's not actually a risk here: `on_pre_build` fires before MkDocs collects `docs_dir` files or builds the nav (confirmed against `mkdocs.plugins.BasePlugin`'s event ordering). I verified this concretely by deleting both files outright (simulating a from-scratch checkout with the untracked/gitignored files just absent) and running `mkdocs build --strict` — it succeeds and regenerates both files before anything needs to resolve them.

## Testing
- `mkdocs build --strict` succeeds with both files present (get overwritten normally).
- `mkdocs build --strict` succeeds with both files deleted entirely (fresh-checkout simulation) — hook recreates them, no nav/build errors.
- `git status` is clean after either build, with no phantom modifications.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ifokkema (approved):
This was an annoying problem for me; thanks for fixing it! LGTM!

reece:
If I recall correctly, my intention here was to regenerate x.y for the active major.minor release, and commit the earlier ones since they will not change. But regeneration is fast so no harm.

